PR #1, opened 2015-04-30
[WIP] First draft of Euler chapter

ketch:
See the notebooks here: 

- http://nbviewer.ipython.org/gist/ketch/445f886de2d09c763f1b
- http://nbviewer.ipython.org/gist/ketch/0a7397cbd3f1d937d1be

This PR combines some existing materials from other places into a first draft of a chapter on the Euler equations.  It covers (in outline, at least) the equations, the exact Riemann solution, and two approximate solvers (HLLE and Roe).  The discussion of the mathematical structure of the equations and of the approximate solvers has not been filled in yet.  This is mostly copied from other notebooks written previously by @rjleveque and myself.

Questions for @rjleveque and @maojrs:

- Does this outline look like the right set of things we want to cover?
- What do you think about the vertical layout of plots?  This requires my branch of Riemann with some modifications to riemann_tools.py; see https://github.com/clawpack/riemann/pull/103.
- I have committed the notebook after clearing output, since I think that makes it easier to use diff.  Is that approach okay with you?
- Do we want static interacts everywhere?  Or, since it is a similarity solution, should we just plot the solution at some particular time?

### DIFF
--- a/Euler_approximate_solvers.ipynb
+++ b/Euler_approximate_solvers.ipynb
@@ -4,6 +4,17 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "# Table of contents\n",
+    "\n",
+    "- [HLLE Solver](#HLLE-Solver)\n",
+    "- [Roe Solver](#Roe-Solver)\n",
+    "- [Comparison of solvers](#Comparison-of-two-approximate-solvers-with-the-exact-solution)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "In the [last notebook](Euler_equations.ipynb) we studied the Euler equations of inviscid, compressible fluid flow -- including the Riemann problem and its solution.  As we saw, the exact solution of the Riemann problem is computationally expensive, since it requires solving a set of nonlinear algebraic equations.  In the context of finite volume methods, the detailed structure of the Riemann solution is almost immediately discarded -- only its impact on the neighboring cell averages is used.  So it makes sense to consider whether we can approximate the solution with less computation.  In this notebook, we investigate approximate solvers for the Euler equations."
    ]
   },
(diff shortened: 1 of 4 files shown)


ketch:
One thing that is missing is plots of the exact solution in state space.  I'm thinking that it would be nice to have something like Figure 14.2 from the FVM book, which uses the p-u plane.  Perhaps this could go to the right of the plots that I have now.

maojrs:
* I like the outline. I think it would be nice to write up the exact solutions for some of the Riemann problems. Should we also do inline implementation of some simple solvers or just refer to the riemann repository? 
* @rjleveque already commented on the vertical layouts.
* Commiting clearing the output is good for me.
* I like the static interacts. They are not very expensive, and the interactivity definitely adds on to the reading experience.  

@ketch @rjleveque should we make a copy of riemann_tools.py into this repository? We can use @ketch latest branch as a starting point.

ketch:
>Should we also do inline implementation of some simple solvers or just refer to the riemann repository?

I think we should provide inline detailed, commented implementations of at least some of them.

> should we make a copy of riemann_tools.py into this repository? We can use @ketch latest branch as a starting point.

We definitely don't want to have two copies.  What would be the motivation for moving it to this repository?

rjleveque:
The motivation would be that we might be actively developing it over the next few months and not want to keep merging into `riemann/master`.   But how about if we create a new branch `riemann/notebooks` for this work in progress?

ketch:
>how about if we create a new branch riemann/notebooks for this work in progress?

That sounds good to me.  Done.

ketch:
BTW, I didn't mean that we shouldn't necessarily copy the file to this repo; just that if we do then we should delete the other version so we don't accidentally work off of it.  The drawback is losing history.

ketch:
Unless there are objections, I'd like to go ahead and merge this PR.  It's still just a draft, but I think it will be easier to finish it after we flesh out some other chapters and especially the introductory material.

The plot formatting issues, of course, can be dealt with in changes to riemann_tools.py, which is independent of these notebooks.

PR #2, opened 2015-05-10
Add overall discretization results to Euler approx solver notebook.

ketch:
When I showed my students the Euler notebooks, they got the mistaken impression that the approximate solvers might give totally wrong answers when used within a mesh-based discretization.  This PR adds results showing that you get pretty good answers from the approximate solvers, and they converge to the right thing as the mesh is refined.

The code could probably be cleaned up some, once we decide on all the things we want to show or emphasize.


ketch:
Updated gist with everything executed here: https://gist.github.com/0a7397cbd3f1d937d1be

PR #4, opened 2015-07-22
Corrected call to riemann_tools.riemann_solution()

MarcKjerland:
Function call riemann_tools.riemann_solution() expects the number of equations (here 3) as an argument, and no longer recognizes keyword argument "verbose." I updated both calls to this function.


ketch:
These changes no longer reflect the API of `riemann_solution()`, so I'm going to close this PR.

PR #13, opened 2017-02-08
Loading libraries without having clawpack

maojrs:
I changed a couple of library calls in case one is running the code without clawpack. If the library is not found, it will still try to load it from clawpack. Note that for JSAnimation to work without clawpack, we would need a copy of the library into this repository, maybe something to consider.


ketch:
There are additional imports from JSanimation further down (in make_html and make_rst), so you might want to do something similar with them.

maojrs:
Wps didn't see those, just changed them too.

rjleveque:
I haven't quite figured out what we are actually using from JSAnimation vs. just the `animation.FuncAnimation` function from `matplotlib`.  Do we really need JSAnimation?

ketch:
Here's a way to produce roughly the same thing with just matplotlib:

http://louistiao.me/posts/notebooks/embedding-matplotlib-animations-in-jupyter-notebooks/

The controls are not as complete -- specifically, you can't step through one frame at a time.  In some situations that's a big disadvantage.

ketch:
@maojrs I think in the future it's a good idea to use separate branches and separate PRs for bug fixes (like the JSAnimation changes) versus adding content (like the Tamman notebook).  That said, for this time it's probably okay to just leave it all here and merge it together.

Should I go ahead and merge this now, or do you expect to add more to the notebook right away?

maojrs:
@ketch thanks for the tip. I am not that familiar with the open-source software development etiquette. Next time I will do bug fixes in a different branch. I will add more stuff to the notebook for sure, but probably not right away, so maybe it is a good idea to merge it now.

PR #14, opened 2017-02-08
Add LWR traffic flow notebook.

ketch:
I included citations to the original LW and R papers.  We could add more citations to recent work, and Randy suggested linking to a java applet but I don't recall where that is.

The notebook also introduces the following concepts:

- Jump conditions (shock speed)
- Similarity solutions
- centered rarefaction waves

At the end, I have included both a widget that lets the user set an left/right states and see the Riemann solution, and a plot like those in other notebooks that shows the characteristics as well as the solution for a fixed pair of states.  We probably don't want to keep both.


rjleveque:
Here's the simulation:
http://www.traffic-simulation.de/index.html

ketch:
Thanks, @rjleveque .  I added a link.  I think it's fine to go ahead and merge this now; it can be polished later.

PR #16, opened 2017-02-14
Extend riemann_solution() to infer stationary waves.

ketch:
In some solvers (such as f-wave Riemann solvers for problems with
spatially-varying flux, as well as many multi-dimensional solvers), there is a
wave with speed zero that is not explicitly calculated or returned from the
solver since it is not needed in the numerical update.  This commit
modifies riemann_tools.riemann_solution() to accept a keyword argument
stationary_wave.  If this argument is set to True, then the states
are calculated inward from both left and right, and a stationary wave
is inferred from the difference between the neighboring states.

If stationary_waves==False, the routine checks is the computed
right state (from summing the waves) is close to the input
right state.  If not, it suggests setting stationary_wave=True.

This commit also adds a fwave option, which divides each wave
by its speed (necessary for output from fwave solvers).

### DIFF
--- a/utils/riemann_tools.py
+++ b/utils/riemann_tools.py
@@ -26,7 +26,8 @@ import copy
 sympy.init_printing(use_latex='mathjax')
 
 
-def riemann_solution(solver,q_l,q_r,aux_l=None,aux_r=None,t=0.2,problem_data=None,verbose=False):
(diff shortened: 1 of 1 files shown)


ketch:
I'm going to go ahead and merge this as I need it for the nonlinear elasticity notebook (upcoming PR).

PR #17, opened 2017-02-15
Add Nonlinear elasticity notebook

ketch:
This notebook shows the exact solution of the Riemann
problem for the nonlinear elasticity system and compares
it with the approximate f-wave solver.

You can quickly see the notebook with cells executed here:

https://gist.github.com/ketch/84aaab7095efb480f740c214262f6a50

However, some of the math doesn't render properly there (it renders fine in a running notebook).


mandli:
I am being lazy about this but the notebook posted at the gist is not rendering completely correctly (the LaTeX, not the image).

ketch:
Yeah, I mentioned that in my original note with the link.  One issue is that underscores are Markdown and I think Github isn't smart about them when they appear inside math.

mandli:
I have had a lot of trouble with course notes I have that do not render properly on github but does locally.  My experience is that it's not necessarily underscores that are the problem but some environments don't render properly.  In particular in this case `\begin{align} \end{align}` does not work well although `$$\begin{aligned} \end{aligned}$$` does.  One of the more annoying ones actually is `~` as `\quad` works instead.

PR #19, opened 2017-02-21
Particle trajectories

rjleveque:
Improve the `riemann_tools.plot_riemann_trajectories` function to allow specifying xmax and density of particles to left and right.

I also added trajectory plots to the Euler and Traffic flow notebooks.

Note:
 - I renamed Traffic\ Flow to Traffic_Flow since my `git diff` doesn't work with spaces in file names
 - I noticed that the `%load` cell in `Euler_equations.ipynb` doesn't get cleared by `nbstripout`, so this has to be cleared by hand before saving notebook.
- Particle paths are not very interesting for Acoustics because the velocity should be scaled by a small number in order for linearized acoustics to be valid (if `p` represents perturbations in pressure multiplied by some large number to make it order 1, then `u` that comes out should be divided by the same large number).  Particle paths using the raw `u` are not physically meaningful and confusing.


ketch:
Maybe we should just remove that `%load` cell, at least for now.  I have found it annoying too (and I'm the one who created it).

rjleveque:
Another option is just `import *` from that file and then include a `FileLink` so those interested can easily click on it to see what they imported.

PR #22, opened 2017-02-23
Add variable speed traffic notebook.

ketch:
Also add citation in basic traffic notebook.


ketch:
I'm going to go ahead an merge this; it's a rough draft that can be revised later.

PR #25, opened 2017-02-26
added Shallow_tracer directory and notebook

rjleveque:
Note:

- This is in a separate directory since I'm using clawpack/classic to compute the solution, so other files are needed.  Should we use pyclaw instead?
- I ran out to time 2.5 with solid wall boundary conditions, so this also shows reflections and interactions, not just the Riemann solution.  I think this is interesting to see with the tracer.
- There's a new Riemann solver in this directory that could be put in the riemann repository instead eventually.

### DIFF
--- a/Shallow_tracer/Makefile
+++ b/Shallow_tracer/Makefile
@@ -1,0 +1,57 @@
+
+# Makefile for Clawpack code in this directory.
+# This version only sets the local files and frequently changed
+# options, and then includes the standard makefile pointed to by CLAWMAKE.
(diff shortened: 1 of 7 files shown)


ketch:
I would be in favor of using Pyclaw. This simplifies the file structure and will allow people to simply "conda install clawpack" and run the examples; no compiler needed. 

I could add a pyclaw version to the same notebook for comparison if that would be helpful.

rjleveque:
That would be great, or better yet copy this notebook up to the top level and pyclawtize it so we can get rid of this subdirectory.

ketch:
I'm closing this since we merged #28.

PR #28, opened 2017-03-02
PyClaw version of shallow water notebook.

ketch:
This is meant as a possible replacement for #25.  It uses PyClaw and avoids the need for a separate directory; it also avoids the need to write or read files.  The content is all the work of @rjleveque; I merely added the PyClaw run and plotting scripts.

Note that for the plotting I did not use visclaw (other than to get IPython_display from JSAnimation).  I think the plots look a bit nicer.  I also took a different approach to the tracer shading, as I am not so proficient with colormaps.  The plotting code is written in a way that can easily be extended to include variable bathymetry.  We might want to move the plotting code to a file.


ketch:
Update: I added the line

    plt.axis('scaled')

so that the area occupied by each tracer region is conserved visually.  This made the plot really skinny so I reduced the depth and extended tfinal.

ketch:
If we decide we want to merge this, I will remove "pyclaw" from the notebook name.  I put it there just to keep track of what is what locally.

rjleveque:
I get
```
ImportError: No module named shallow_roe_tracer
```

ketch:
Ack. Git add failure. I will fix it later tonight when I'm at my laptop.

ketch:
Okay, it should work now.  By the way, I amended and force pushed (sorry), so you will probably need to just delete my old branch and check this out.

ketch:
Okay, I got rid of the gaps.  Were you planning on adding more to this notebook?  If so, maybe I should close this PR and instead submit this as a PR against your branch; then you can merge and keep going.  Or feel free to just take this and open your own new PR with additional content.

ketch:
After discussion with @rjleveque I'm going to merge this and we can expand it in the future.

PR #33, opened 2017-03-04
Add calls to `plt.show()` in various places.

ketch:
This seems to be necessary after recent updates to
matplotlib and/or ipywidgets.

@rjleveque Please check whether this breaks the widgets when combined with your current versions of things.


rjleveque:
Works fine with ipywidgets version 5.2.2 on Jupyter 4.2.0.

Issue #32, opened 2017-03-04
Tamman EOS notebook generates errors


ketch:
@maojrs I thought I successfully ran this notebook back when it was added, but I must be wrong since it doesn't work for me now and it hasn't been changed.

The second code cell gives me an error about passing 5 arguments to  `phi_exact`.  I was able to fix that, but then I get an error about `ul` being undefined.  Does the notebook run for you?

maojrs:
It does not really matte since I just did a major update of the notebook. Now it uses riemann_tools,  and I also tried to follow a similar approach than the one used with the ideal gas Euler eqs. I am still working on it, but I will make a pull request in case you want to merge it.

PR #34, opened 2017-03-04
Add test script, requirements, and travis script.

ketch:
The test script simply executes each notebook.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: python
+sudo: true
+
+python:
+  - "2.7"
+  - "3.6"
+
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq gfortran
+  - git clone --branch=master --depth=100 --quiet git://github.com/clawpack/clawpack
+  - cd clawpack
+  - git submodule init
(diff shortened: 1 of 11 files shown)


ketch:
This tests all notebooks using both Python 2 and Python 3.  Right now the tests fail because of #32.  I'm planning to merge this anyway, but we should fix #32 ASAP and then avoid merging anything that breaks the tests.

I also added installation instructions to the README, which should give a user all the dependencies.  Anyone can run the tests via

    python test.py

Note that we don't actually check that the output is correct; we only check that no exceptions are raised.

ketch:
By the way, I stole the code for testing from http://blog.thedataincubator.com/2016/06/testing-jupyter-notebooks/.  I had to modify it to make it work with Python 2 and 3.

ketch:
I'm going to make this work with nosetests and make it run the doctests.

ketch:
Okay, now nosetests will pick this up.  You should run it with

    nosetests --with-doctest

to run the doctests too.  Also,

   python test.py

still works fine.

This is ready to merge.

rjleveque:
It passes also on the Tamman notebook after I merged in #35 on my laptop.

PR #36, opened 2017-03-05
Color code waves

ketch:
Shocks are red, rarefactions blue, contacts (and stationary
interface jumps) are black.  This applies to the characteristic
rays in the x-t plane as well as the hugoniot loci and integral
curves in the phase plane.

This required a major change to the exact Riemann solver design and interface.
Exact Riemann solvers should return an additional list of wave types,
with one entry for each wave and each entry set to one of
{'shock' | 'raref' | 'contact'}.

For non-exact solvers, since everything is just approximated by jumps
and we don't know what kind of wave it really is, we just color
everything black.

The changes incorporated here will necessitate minor changes also in #35 and #25 or #28,
since those also include exact Riemann solvers.


ketch:
I'm going to go ahead and merge this so that I can help make #35 and #28 compatible with it so we can get to a state of tests passing.

PR #38, opened 2017-03-06
Euler with Tammann EOS

ketch:
This is equivalent to #35.  I'm just issuing a new PR that includes my updates to get Travis to run the tests.


ketch:
Merging this so that we are in a "tests passing" state.

PR #42, opened 2017-03-13
Added interactive pplane html apps

maojrs:
Interactive apps, added two version of each (acoustics and shallow water), one big and one small for compact display.


rjleveque:
I don't see any difference between the big and small versions, and they seem to be the same size file:

```
-rw-r--r--  1 rjl  staff   106K Mar 13 22:36 phase_plane_acoustics.html
-rw-r--r--  1 rjl  staff   106K Mar 13 22:36 phase_plane_acoustics_small.html
-rw-r--r--  1 rjl  staff   281K Mar 13 22:36 phase_plane_shallow_water.html
-rw-r--r--  1 rjl  staff   280K Mar 13 22:36 phase_plane_shallow_water_small.html
```

maojrs:
They are basically identical except one displays at a smaller size by default when opening in the browser. If we do not need them, I can remove the small versions.

rjleveque:
Oh I see, it's a subtle difference and the small one is still pretty big.  Maybe have one even smaller?

maojrs:
Added two even smaller versions of the pplane interctive app.

ketch:
These look great.  Can either of you remind me how to embed them in a notebook?

ketch:
Never mind, I just remembered.  (%%html)

ketch:
Okay, better question: do we know how to display these things inside a notebook without pasting all the code into a notebook cell?  So far I have:

1. Pasted the contents of the html file into a notebook cell, with `%%html` at the top.  This works fine, but is very ugly.
2. Put the following in a notebook cell:

```python
from IPython.core.display import display, HTML
app = open('phase_plane_shallow_water_small.html','r').read()
display(HTML(app))
```
This only shows a static view; the points are not draggable.

maojrs:
I tried:
from IPython.core.display import display, HTML
app = open('phase_plane_shallow_water_small.html','r').read()
display(HTML(app))
 
and it showed the interactive app with no problem, the points were draggable,
and everything worked as expected.

Another possibility would be to do it as I do it as I did in this notebook:
https://github.com/maojrs/ipynotebooks/blob/master/interactive_test.ipynb
However, that requires having mpld3 installed.

ketch:
@maojrs That's strange.  What browser are you using?  What version of Jupyter?

maojrs:
I am using chrome with Jupyter 4.4.1. What about you?

rjleveque:
You can just do this in a markdown cell:

```
See this [html file](phase_plane_shallow_water_small.html)
```

Dragging works fine.

ketch:
@rjleveque Thanks!

ketch:
> I am using chrome with Jupyter 4.4.1. What about you?

I'm also using chrome with Jupyter 4.4.1.  I have noticed that after I run the cell, the dragging doesn't work.  But if I then refresh the page it does.  Strange.

maojrs:
Well then maybe it is better just to do the markdown version, to avoid this happening to other people.

maojrs:
Oh I just noticed the markdown version doesn't do it inline. I think inline would be better. For me it works the first time without need to refresh...

PR #43, opened 2017-03-14
Tammann nb

maojrs:
interactive_phase_plane
Updated version of the Tammann notebook; I still need to add approximate solutions. However this update implements an interactive phase plane which could be easily extended to work with other nonlinear systems. Before working more on it, I want to know  how well it works for others.


rjleveque:
Really nice!  I like the way you can turn the unphysical solution off to see the physical solution better.

ketch:
This is great!

I have issued a PR against your branch with some minor spelling and grammar corrections that I made while reading it over.  There were a lot of inconsistencies in capitalization of variables and subscripts; I probably didn't catch all of those!  I also didn't check for mathematical correctness.

Regarding the widget: this is amazing.  I have wanted to put sliders side-by-side for a long time.  Is it possible to put the widget controls next to the plot (instead of above)?  Vertical space is our most scarce resource.

Here are a few other comments; we don't necessarily need to resolve them before merging this:

- The initial description of how to solve the Euler Riemann problem applies to ideal gas Euler as well, so perhaps it should appear earlier (or perhaps we can leave it here and use this chapter to discuss in detail how to solve the Riemann problem exactly)
- Perhaps we should use $s^1, s^2, s^3$ for the wave speeds here?
- In this notebook subscripts "_*" are used for intermediate states.  In other notebooks we have used superscripts.  It might be good to use something consistent throughout.  I prefer superscripts; even in the current notebook, the subscripts collide with other uses of subscripts.
- I find it confusing to use $Q$ to denote a flux.  Maybe $F$?
- In some equations we have $\mathcal{Q}$, but I don't think that has been defined.
- I was confused by what's written in the very last cell.  Is this accidentally left over from an earlier version?
- It would be nice to add some examples showing how this EOS gives very different results from ideal gas Euler.
- Regarding equation numbers: I notice you're using a very verbose approach with \tag{}.  Is there a reason for this?  I've been using just regular latex \label{} and \ref{}.

One last thing: if you don't mind updating to the latest version of jupyter (`pip install -U jupyter`) we'll all be on the same version and have less noise in the commit diffs.  It seems they added more cell metadata in a recent release.

maojrs:
Thanks for spotting some of the notation inconsistencies. I have changed the notation of this text so many times that it is hard to keep track. 
- Regarding sliders on the side of the plot, I am not sure if it is possible. I will take a look and see what I can do. There might be other solutions that use less vertical space. 
- I think the description of how to solve the exact Euler equation should go in the ideal gas Euler eqs part (or even in the shallow water). Of course, this would have to be updated to mention only how to derive the RH-conditions and integral curves for this particular case.
- About to notation, I don't really mind which one we use. I will change the speeds to be s_1, s_2 and s_3. - Mathcal Q is a leftover of some previous version, I still have to do a consistency check to make sure everything matches.
- The last cell is just a reminder for myself of some things I have pending in the notebook. It is not yet a final version, but I wanted to show you the phase plane app.
- Regarding examples, I have a couple that are relevant for the Tammann EOS. I am planning on adding them soon.
- About the equation numbers, I use tag, so the numbering works in the notebook. Is "label and ref" enough for the references to work in the NB?

Just updated jupyter :)

ketch:
>Regarding sliders on the side of the plot, I am not sure if it is possible. I will take a look and see what I can do. There might be other solutions that use less vertical space.

Okay.  In any case, what you have is very nice.

>I think the description of how to solve the exact Euler equation should go in the ideal gas Euler eqs part (or even in the shallow water). Of course, this would have to be updated to mention only how to derive the RH-conditions and integral curves for this particular case.

Agreed.  Let's keep it here for now with the plan to migrate it as we coordinate chapters later.

>About the equation numbers, I use tag, so the numbering works in the notebook. Is "label and ref" enough for the references to work in the NB?

Ordinary labels will work if you put this in a cell at the top of the notebook:

```
%%javascript
MathJax.Hub.Config({
  TeX: { equationNumbers: { autoNumber: "AMS" } }
});
```

That is a bit annoying but I imagine there is a way to hide the cell.

ketch:
@maojrs I think your latest updates include some unintended things:

- You've modified lots of other files outside the Tammann notebook
- You've undone the grammatical corrections from my earlier PR

maojrs:
@ketch I will double check since I had to move everything to my old computer (new one was kaput), 
and that might have mixed up things. Some of the file changes to other files were intentional; I added the interactive pplane for the Euler equations (ideal gas). In the last commit I move all these changes to a new file interactive_pplane.py. The changes in Euler.py reflect only a test example.

maojrs:
@ketch Ok, manage to fix this up, sorry for the mess. The history should be clean now. Please let me know if you find any issues. The latest commit reflects my most recent modifications. I added interactive_pplanes.py to store the code to generate interactive phase planes. There is a new version now for Euler eqs with ideal gas EOS. A test example of this one is shown at the end of the Euler Tammann notebook just for reference (maybe it should be moved to the ideal gas Euler notebook).

ketch:
Looks like you fixed it.

A couple of other things I noticed:
- It looks like you are escaping all your backslashes in math mode (using `\\` in place of `\`).  I would prefer not to do this because it will cause issues when we try to generate the print book.  I know that Github may not render things correctly, but I think that is less important (Github will choke on our labels anyway, I think).
- As discussed by skype, let's stick to number indices for anything related to the waves.  Thus the speeds can simply be $s_1, s_2, s_3$.

I will take a closer look and add more if I have other comments.

ketch:
@maojrs Just three more comments:

1. This is really looking nice!
2. We can drop the javascript cell at the top, now that we are using the nbextension for equation numbers.
3. I'd like to refactor your interactive phase plane code in order to facilitate a lot more code reuse and make it easier to extend this to other systems.  But I don't want to mess you up if you're currently modifying that code.  Let me know if it's okay to do some heavy editing of `interactive_pplanes.py`.

maojrs:
@ketch it took me a while to reply since I've been tied up with other things. 

-I am not sure what you mean with the backlashes. I use \\ in math mode to break a line,
\ would only make horizontal space. This should be fine in the latex version, or am I missing something?

- I was going to change the velocities to s_1, s_2 and s_3, but I do a general derivation which uses s_k and q_k in the same equation,  so it makes sense to use k=l,r instead of k=1, 3. It could be changed to numbers, but I think it would be a bit confusing to use different indexes for q and s. What do you think?

- Javascript cell dropped!

- I was doing the same (generalizing the interactive_pplane code), but then I realized there are not that many systems we want to apply this two, and it might be better to have an individual code for each. The main reason I am thinking this is code readability and location of widgets within the GUI (some systems have more parameters, different relevant plot ranges and so on). This is why I decided to move everything into the interactive_pplane.py script. One possible idea would be to create a quite general parent class that can be used with an arbitrary system, but then we can make a child class to really modify the interactive GUI of the systems we really want to present in a very neat way. I can give a shot to this. What do you think?

PS: Just modified the interactive app so now we have tabs and more vertical space than before.

ketch:
I went over the code just to familiarize myself with it, and did some cleanup as I went.  I've issued a PR against your branch.

>I am not sure what you mean with the backlashes. I use \ in math mode to break a line,
\ would only make horizontal space. This should be fine in the latex version, or am I missing something?

My mistake - never mind!  However, there is one other thing that I think will break when converting to latex: in a few places, you use both $$ and \begin{align} around equations.  I suggest using only one or the other.

>I was going to change the velocities to s_1, s_2 and s_3, but I do a general derivation which uses s_k and q_k in the same equation, so it makes sense to use k=l,r instead of k=1, 3. It could be changed to numbers, but I think it would be a bit confusing to use different indexes for q and s. What do you think?

I don't think this is wise.  The state $q_l$ is always to the left of $x=0$, but the wave with speed $s_1$ could be moving to the right.  Using $s_l$ to denote the speed of a right-moving wave seems like it creates confusion.  In fact, I'm not sure that your description in the notebook is correct for supersonic flows; have you checked it?

> PS: Just modified the interactive app so now we have tabs and more vertical space than before.

That looks great!  I would suggest combining the left and right state tabs; I've done this in my PR.  Since other tabs already have more than one row of widgets, we don't lose any vertical space, and the user doesn't have to switch tabs as much.

ketch:
@maojrs If it's okay with you I'd like to go ahead and merge this.  I'm planning to rework the Euler ideal gas notebook and it would make sense to use what you've done here.  The notational issues can be worked out in the future.

maojrs:
Great! Any other changes I will push them to a new branch.

PR #46, opened 2017-03-21
Initial draft of kitchen sink chapter.

ketch:
This chapter investigates the circular hydraulic jump.  It is different from the other chapters since the states away from the jump are smoothly varying rather than constant.

One issue at the moment is that you have to execute the javascript cell to get equation numbering working.  I believe I can fix this soon.

### DIFF
--- a/Kitchen_sink_problem.ipynb
+++ b/Kitchen_sink_problem.ipynb
@@ -1,0 +1,902 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "import numpy as np\n",
+    "from scipy.optimize import fsolve\n",
+    "from scipy import integrate\n",
+    "import matplotlib.pyplot as plt\n",
+    "from clawpack import pyclaw\n",
+    "from clawpack import riemann\n",
+    "from clawpack.visclaw.ianimate import ianimate\n",
+    "import matplotlib\n",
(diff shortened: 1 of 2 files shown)


ketch:
Okay, I removed the javascript cell for equation numbering.  Instead, you should install the equation-numbering extension.  In a terminal, do this:

    pip install jupyter_contrib_nbextensions
    jupyter contrib nbextension install --user
    jupyter nbextension enable equation-numbering/main

You may notice that when re-rendering markdown cells with numbered equations, MathJax fails to render them.  Just push the new toolbar button (at the right end of the toolbar) and the rendering + numbering will refresh.

rjleveque:
Nice! 

I haven't gone through in detail yet, but I noticed a warning message:

```
r_jump, r, profile = jump_location(Q=200.,r_jet=1.,h_inf=1.,g=980.,tol=1.e-6)
print('Jump radius: '+str(r_jump)+' cm')
plt.clf()
plt.plot(r,profile)
plt.xlim(r[0],10);

/usr/local/lib/python2.7/site-packages/scipy/integrate/odepack.py:218: 
ODEintWarning: Excess work done on this call (perhaps wrong Dfun type). 
Run with full_output = 1 to get quantitative information.
warnings.warn(warning_msg, ODEintWarning)
```

ketch:
> I noticed a warning message

Yes, it's annoying -- the ODE solution approaches zero when integrating inward, and this gives the integrators a lot of trouble.  This occurs in a region that we don't care about (it's inside the jump radius, and this is the part of the solution that we use only outside the jump radius).  I could do something special to get around it but since it doesn't give the wrong answer I haven't bothered.  It is annoying to have a warning appear in something that's supposed to be polished, though.

ketch:
I'll go ahead and merge this just so we can keep it updated and working with everything else.

PR #47, opened 2017-03-22
Initial draft of advection notebook.

ketch:
I suggested yesterday to @rjleveque that maybe we didn't need this, but after thinking about traffic flow and other physical applications I thought that it would be worth introducing this just to get the idea of conservation of mass out of the way (not that we want to spend much time on that in this book), and it wouldn't hurt to have this basic, trivial example of a Riemann problem.  I'm still wondering if we want to place it before or after the acoustics chapter...

### DIFF
--- a/Advection.ipynb
+++ b/Advection.ipynb
@@ -1,0 +1,205 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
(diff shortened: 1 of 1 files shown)


rjleveque:
@ketch: The figures are missing,

```
![Characteristics in the x-t plane](./figures/characteristics.png)

![Propagation along characteristics](./figures/char_prop.png)
```

PR #49, opened 2017-03-27
Shallow water notebook

ketch:
This material is mostly cannibalized from existing material authored
by @rjleveque.  The notebook has a lot of content but is missing
several pieces, including much of the exposition and the interactive
phase plane.


ketch:
To be clear, this is not really ready for review yet.  I'll post a request when it is.

ketch:
Okay, this is ready for review now.

I struggled with how to organize things, because most of this is covered in FVMHP chapter 13, and the presentation there seems to have been perfected.  I have followed much of that.  The main difference is that I present much less detail, due to the medium we're using, and also because the reader can refer to that text if desired.  Obviously we are adding to that presentation in terms of including the code and making the plots interactive.  I also derived the centered rarefaction wave structure in a way that I feel is simpler (though perhaps less insightful).

I'm sure there is plenty that could be improved, so please take a look and post comments.

rjleveque:
Nice! Some comments...

1. I like the ability to turn on/off plotting the characteristics.  But when the 1-characteristics are shown in the first place this is done (the all-shock solution), it shows them in the middle state and the right state whereas it would be more useful to plot them in the left and middle states.  Similarly the 2-characteristics should be plotted in the middle and right states.

2. Why are there gaps in the plots of the integral curves?

3. In the all-rarefaction solution of the dam-break problem, it would be nice if we could plot the overturning wave as in Figure 13.16 of FVMHP.

4. Right after the `Link to phase plane demo` perhaps note that the left and right states can be dragged around.  People don't always realize that when I've pointed them to this before.

5. The final cell seems ok to me, but I don't know that it adds much as long as the html demo page works for people.

ketch:
>I like the ability to turn on/off plotting the characteristics. But when the 1-characteristics are shown in the first place this is done (the all-shock solution), it shows them in the middle state and the right state whereas it would be more useful to plot them in the left and middle states. Similarly the 2-characteristics should be plotted in the middle and right states.

Agreed.  This is actually due to bugs in the characteristic plotting.  I've been planning to deal with that but haven't gotten to it yet.

> Why are there gaps in the plots of the integral curves?

This is an annoying quirk of matplotlib's streamplot function (MATLAB does it too, I believe).  It stops tracing a streamline sometimes when it gets too close to another.  I wrote the code and text for this before I discovered this issue, and I decided to leave it in there to see how much it bothered you.  :-)

I know how to "fix" this feature and I've proposed it to the matplotlib devs: https://github.com/matplotlib/matplotlib/issues/8388.  That won't happen in time for our book proposal though.  I could just write some code manually that integrates a few integral curves -- do you think that would be useful?

> In the all-rarefaction solution of the dam-break problem, it would be nice if we could plot the overturning wave as in Figure 13.16 of FVMHP.

Okay.  I wasn't sure how closely we wanted to follow FVMHP, but I guess we have your permission. ;-)  A patch would be welcome, or I will see about doing it.

> Right after the Link to phase plane demo perhaps note that the left and right states can be dragged around. People don't always realize that when I've pointed them to this before.

Good idea; I'll add that.

>The final cell seems ok to me, but I don't know that it adds much as long as the html demo page works for people.

That was my feeling too.  Perhaps one nice thing about it is that you can directly set the velocity, which is much more natural (to me) than setting the momentum.

maojrs:
Great! A couple of extra comments:

When I click on the plot_1_chars or plot_2_chars widget I get an error:
"invalid index to scalar variable"
Not sure if this is happening to anyone else.

To avoid the gaps in the streamplot, maybe we can just write our own 
plotting function manually, just an idea.

It seems to me the time interactive plots in the horizontal format look a 
little bit cluttered, especially with 4 plots in one line. As we are including
the phase plane plot, maybe a 2x2 plot grid could look better? Other possibility
would be to plot the phase plane in another cell.

rjleveque:
@ketch: 

I'll see if I can do the overturning waves.

Regarding setting velocity rather than momentum, we could also consider adding phase plane plots in the h-u plane (rather than h-hu).  These are also good for showing how a dry state arises in the Riemann solution with symmetric outflow, so something I should have included in the book but didn't.  I'll try to do something...

@maojrs:

I didn't have any problem with clicking the widgets (on Chrome).

I think 4 plots in a row is ok for the notebook because it fits in the screen better than an array.

ketch:
I've added a new method for plotting the characteristics, following a suggestion from @rjleveque .  It's still not perfect, but it's a big improvement.  The main issues now are:

- Characteristics accumulate at a shock, resulting in a noticeable black line peeking out from under the red one.  Perhaps this is a feature.
- Only characteristics starting from the t=0 axis are traced.  Sometimes there are regions in the Riemann solution with few or zero characteristics that cross this line, so those regions don't show any characteristics.

I also added an explanation of the draggable points in the javascript phase plane.  Maybe we want to put a note about them in that html page itself.

ketch:
> Regarding setting velocity rather than momentum, we could also consider adding phase plane plots in the h-u plane (rather than h-hu). These are also good for showing how a dry state arises in the Riemann solution with symmetric outflow, so something I should have included in the book but didn't. I'll try to do something...

We could easily switch all the phase plane plots to use the h-u plane if we decide that's better.  I don't have a good sense of which is more useful.

ketch:
I just updated this to include a section on dry states.  I also changed a lot of it to work with the h-u plane rather than h-hu.  I believe the exact Riemann solver for SW now works for all possible inputs with non-negative depth; this includes dry left/right states as well as dry middle states.  Please try your best to break it.  Other improvements:

- Characteristics and other plotting works with dry states
- depth plots are done with a blue fill to suggest water
- probably more I have forgotten

I'm pretty happy with this notebook now, feature-wise.  I'm sure it can be improved still in pedagogical terms.

rjleveque:
Looks great, thanks for all the work on this!

PR #58, opened 2017-04-06
[WIP] Initial draft of Prologue.

ketch:
This is a very preliminary draft of the zeroth chapter, following notes at https://github.com/clawpack/riemann_book/wiki/Introductory-notebook-outline.  It's pretty straightforward, but I'm not very happy with it.  It feels too dry for an introduction.  I'm wondering if we should turn it on its head by starting with an interesting example as motivation and then working backward to the basics from that.

I haven't filled in the later parts of the chapter (shallow water example, approximate solutions, and generalizations).

Please criticize, or just take this and rewrite it in a better way.


ketch:
Closed in favor of #62.

PR #60, opened 2017-04-08
Add references and notes on large Froude number limit in pressureless flow chapter

ketch:


ketch:
Warning: I deleted a commit from this branch and then pushed another.  If you had already fetched it, you'll want to just delete the old version and refetch it.

PR #61, opened 2017-04-10
Modified reval function for shallow water to return nan in overturning case

rjleveque:
So that it plots a gap rather than a discontinuity that doesn't agree with the x-t plane plot in the `Shallow_water.ipynb` notebook.

### DIFF
--- a/Shallow_water.ipynb
+++ b/Shallow_water.ipynb
@@ -532,7 +532,10 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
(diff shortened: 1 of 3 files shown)


ketch:
Thanks!  I wasn't sure what to plot in that region, but I think leaving it blank makes sense.  I will remove some additional code that I had added just to modify the plotting in that region (since it's now unnecessary).

ketch:
Actually, this looks good as it is.  I'll go ahead and merge.

PR #62, opened 2017-04-11
Another possible prologue

rjleveque:
I tried starting a prologue based on @ketch's suggestion in #58 that we might want to start first with an interesting example.  It's in `Prologue2.ipynb` and this PR includes also @ketch's `Prologue.ipynb`.


ketch:
I like this and think it is a nicer approach.

maojrs:
Just took a look at this version. I really liked starting with the examples and animations. For the animations, what are you thinking? JSAnimations or perhaps interacts with a time slider and without the x-t plane?

rjleveque:
I suggest closing this in favor of #68, in which I combined these Prologue notebooks into a new `Introduction.ipynb` file that uses the SW examples with colored stripes.  

I think we should call it *Introduction* rather than *Prologue* since it's really the introduction to some of the basic concepts of Riemann problems and everyone should read it before going on to other notebooks.  

We might also want a Prologue, but I think that might have more basic information about how to use Jupyter notebooks, where to find things online, etc.

PR #63, opened 2017-04-16
Rewrite exposition of variable-speed traffic notebook

ketch:
This rewrite is based on the viewpoint that all the solutions with
transonic rarefactions (including those that also have a shock)
include the stationary jump as well as composite wave that is
resonant with the stationary one.

It's much clearer now why we end up with the specific solutions
that we do.

This also replaces q with rho in both traffic notebooks.


ketch:
This is ready to be merged now.

rjleveque:
These notebooks look great to me!  A couple minor suggestions on the notebook `Traffic_flow`:

When discussing the Green Light, you say it's called a rarefaction wave by analogy to fluid dynamics, but it might be better to say it's called a rarefaction wave "because the drivers experience a decrease in density, or rarefaction, as they pass through this wave".  I think rarefaction means decrease in density more generally than just gas dynamics.

`traffic_LWR.plot_car_trajectories` doesn't seem to plot a horizontal line at the value of `t` set by the slider the way we have in other plots in the x-t plane. Also the vertical axis is mis-labeled as `q` and should be `t`.

The last section on `Riemann solution with characteristics` doesn't seem to show any characteristics, or have any discussion.

The notebook `Traffic_variable_speed` has a lot of great examples! Thanks for figuring all this out.

ketch:
Thanks @rjleveque for the helpful comments.  I've made all the changes/corrections you suggested, and merged that last section of the notebook into the interactive widget before it, so now it includes characteristics and trajectories.

PR #64, opened 2017-04-16
WIP: Possible approach to show static view of widget-generated plots

rjleveque:
Here's one way we could create a static view of plots made by widgets: Create a new `interact` function that just produces a single plot using the `value` attribute of any widgets specified.

The output can be viewed at
https://nbviewer.jupyter.org/url/faculty.washington.edu/rjl/Sample_Euler_equations.ipynb


ketch:
That seems like a very convenient way to handle this for now.  I would vote for putting the "fake" interact function down in utils.  For automatically generating all the notebooks with static output, we could just have a script that replaces the interact import line in the JSON file of each notebook and then executes it.

rjleveque:
@ketch: I moved this function to `utils/interact` and for now have a simple cell to select which version to use:

```
#from ipywidgets import interact        # for interactive widgets
from utils.interact import interact     # for static figure that can be viewed online
```

ketch:
I'm happy to merge this now, unless you are still planning to develop it in some way.

rjleveque:
Should this remain in a separate file `utils/interact.py` or should we move it to `utils/animation_tools.py`?  It is somewhat different but related. 

If we keep it in a separate file should we choose a different name than `interact` since it's potentially confusing to have the same name for module and function?  Maybe `snapshot_interact.py`?

ketch:
I would leave it in a separate file for now, since it's helpful to keep that top-level print statement as a reminder to us.  It might make sense to include it in animation tools when we merge that back into Clawpack, since I could imagine it being useful in other contexts.

Changing the filename to `snapshot_interact.py` seems like a good idea since it gives a better indication of what it really does.

rjleveque:
I decided to rename the module as `utils/snapshot_widgets.py` since it's used instead of `ipywidgets`.

I tested on a couple other notebooks and added to `Pressureless_flow` and `Traffic_variable_speed`,  with the default being to use `ipywidgets` so we should change it if we explicitly want to make a snapshot for posting.  

Perhaps we can create a way to do this automatically if the notebooks are executed via `nbconvert`.

In `Traffic_variable_speed` I had to change the last couple cells to use `t = FloatSlider(...,value=0.5)`  in order for `snapshot_widgets.interact` to work, it fails with the older style `t=(0,1,0.1)` since it uses `value`.

Eventually the cell 
```
from ipywidgets import interact                   # for interactive widgets
#from utils.snapshot_widgets import interact      # for static figure that can be viewed online
```
can be added to other notebooks, but for now I think we can merge this any time unless there are further suggestions.

rjleveque:
I'm going to merge this but feel free to make further suggestions.

PR #65, opened 2017-04-17
Set t_pointer to False by default.

ketch:


rjleveque:
This is fine for now. Maybe we can improve it at a later date.

PR #68, opened 2017-04-23
WIP: make plots with red/blue stripes advected with flow

rjleveque:
In the past I've made animations of shallow water with advected colors only by running Clawpack.

Now I'm working on a tool to do this from the exact Riemann solution. I separated out the code to compute trajectories from the previous code to plot trajectories in the x-t plane in `utils/riemann_tools.py`, and use it to compute the motion of the edges of the stripes.

So far this is only illustrated in `Euler_equations.ipynb`, where density is colored red to left and blue to right of contact discontinuity.

I plan to use this in the Prologue notebook for the examples of shallow water equations.

I also plan to extract more of the messy code now in `Euler_equations.ipynb` and put it in `riemann_tools`, but welcome comments or suggestions before doing more.


rjleveque:
I added an `Introduction.ipynb` notebook, see #62.  

We might want to improve this to allow JSAnimation as an option instead of the widgets so that people viewing the introduction online can see the time evolution.

ketch:
Looks good!  Moving the stripe plotting code to `riemann_utils` and incorporating it as a standard option with `make_plot_function` seems like a good idea.

I wrote a JSAnimation version of the stripe plotting in Shallow_tracer.ipynb.  It avoids rasterization of the figures and looks just as nice as the widget version.  However, it uses a different approach; it replaces the contents of the plot objects rather than creating a new plot each time.  I think this is a preferable approach in general, but it takes a bit more work and it's different from most of our code for the book.  The problem here is that the inline matplotlib backend and the JSAnimation code expect different things.

For the links in the introduction, it might be nice to refer to chapters rather then filenames; e.g.

    [The chapter on shallow water](Shallow_water.ipynb)

rather than

    [Shallow_water.ipynb](Shallow_water.ipynb)

ketch:
One other issue: personally, I think it would be better if these widgets started with t=0.  But for the print book we'll want to use a non-zero time.

rjleveque:
@ketch: I changed to `t=0` but your modification of `snapshot_widgets.py` in #69 doesn't reset to `t=0.2` because elsewhere `t=0` is converted to `t=0.001` to avoid a divide by zero at the initial time.

rjleveque:
This is rather clumsy but I managed to make a version of `interact` that creates a JSAnimation instead of the interactive widget.  I had to introduce a `fig` parameter to get this to work properly for all three styles of `interact` since for JSAnimation it has to accumulate a list of figures.

It is illustrated in `Introduction.ipynb`.

maojrs:
Just tried the interacts with JSAnimation; they work quite well for me. Nothing clumsy.

rjleveque:
What's clumsy is that the function `exact_solvers/shallow_water.plot_shallow_water_demo` has to take a parameter `fig` and uses this to determine whether to show the figures or not depending on whether JSAnimation is being used, and then in the `Introduction.ipynb` notebook `fig=ipywidgets.fixed(0)` is passed in so that it doesn't make a widget for the parameter `fig`.

So this `jsanimate.interact` can't be used in other notebooks currently or it produces plots at all output times, not what we want.  Making it work more generally would require modifying a bunch of our other functions that make plot functions and corresponding calls to `animate`.  

For now at least I think it's ok to just have the jsanimate version in the Introduction, and then we can discuss later whether to try to make it work more generally.

rjleveque:
@ketch: do you understand why Travis is failing?

rjleveque:
I suggest closing this in favor of #71.

PR #69, opened 2017-04-24
Initial effort at compiling book.

ketch:
To compile the book, you’ll need to update to the latest pandoc (1.19.2) from
https://github.com/jgm/pandoc/releases/tag/1.19.2.1 and install takluyver’s
bookbook (with pip).  I think you also need to have Clawpack installed for
Python 3.

Once you have all that, run the shell script:

    ./make_book.sh

The script generates a bunch of new notebook files with numbers in front of
them, and those are used to compile the book.  You can remove the extra
notebooks afterward by doing

    rm 0?-*.ipynb

The set of notebooks to be included, and their order, is determined by the list
at the beginning of `make_chapters.py`.  For the moment I’ve just included
three sample chapters.

Cells can be omitted (hidden) in the book output.  For instance, the cells with
import statements have been hidden.  To set a cell to be hidden, open the
notebook and select view -> cell toolbar -> tags.  Then add the tag ‘hide’ to
the desired cell(s).

To use the static versions of the widgets, I’m just doing a search and replace.
This means that the following code should appear by itself on some line of the
notebook:

    from ipywidgets import interact

bookbook expects that each notebook begins with a Markdown cell containing a
level-1 header, which will be the chapter title.

I removed the messages printed by snapshot_widgets because they are a bit of a
nuisance in the book.

For convenience, a copy of the pdf is attached to this message.

[riemann.pdf](https://github.com/clawpack/riemann_book/files/951703/riemann.pdf)


rjleveque:
The pdf looks great, thanks for getting this going!  

I'll add a few comments later and try out the scripts...

ketch:
I added more chapters.  If you're experimenting with changing the formatting of the book, it's probably convenient to edit `make_chapters.py` to include just 1 chapter so that it compiles more quickly.

Here's a PDF with 8 chapters and about 100 pages: [riemann.pdf](https://github.com/clawpack/riemann_book/files/951918/riemann.pdf)

This is not all the chapters we have drafted, but I think it includes all the ones that are in reasonably polished shape (and some that perhaps are not).

rjleveque:
I suggest closing this in favor of #71.

ketch:
Could we just merge this and then merge #71?  Then there will be a less massive set of changes to review there.

ketch:
And yes, I'd like to go ahead and merge this.

PR #70, opened 2017-04-24
Added Tammann NB examples

maojrs:
Just an update on the tammann Euler NB. Now it has examples.


ketch:
Nice!  I issued a PR with some small modifications intended to make it easier to read the code.

It would be nice to include plotting of the characteristics as is done now for instance in the Euler and shallow water notebooks.

Regarding the last example: the text implies that the negative pressure is somehow caused by the Tammann EOS.  But I believe the exact solution to that Riemann problem exists and has non-negative pressure, so really the negative pressure there is just because we haven't written the exact Riemann solver to correctly deal with vacuum states.  Right?

maojrs:
Great! Incorporated changes and made a couple of minor changes myself. You are right; the negative pressure states are mathematically correct. However, I wanted to make a point that this can happen quite easily with the Tammann EOS. I rewrote a couple of sentences in this part to make that clear.

By plotting of the characteristics, you mean the plot of the "plot particle trajectories" section of the Euler NB?

ketch:
> You are right; the negative pressure states are mathematically correct.

Actually that's not what I meant to say.  I'm not sure I like the idea of talking about "mathematically correct" values for a model that has obviously been taking outside the regime where it is valid.  My inclination would be to either

1. Correctly code the solver to handle vacuum states; I'm not sure how difficult this is but it was surprisingly simple for the ideal gas case.
2. Or omit the example and just mention that additional care is needed when the solution leads to a vacuum state.

>By plotting of the characteristics, you mean the plot of the "plot particle trajectories" section of the Euler NB?

No, I mean the characteristics that are plotted along with the waves (shocks/rarefactions) in the regular Riemann solution plots.  If you haven't looked at the notebook for awhile then you won't have seen them.  Update master and run the cells; I think it will be obvious what I'm referring to.

maojrs:
I will try to incorporate the vacuum states into the solution then. I see now, I missed the characteristics dropdown menu, will try to get those too.

maojrs:
Actually it was more complicated than of what I thought. In the Tammann EOS you can have negative pressures without being a vacuum state. Solving the Riemann problem for vacuum states is actually not trivial because the vacuum state pressure that yields zero density is p=-pinf, and if these are different across the interface, you would have two different values of p in the middle state, so I don't really know how to solve it. Nonetheless, I added a section to explain the negative pressure cases, and I mentioned the possibility of finding vacuum states. Hopefully it is clear. I also added the characteristics into the plots. If you get a chance, let me know what you think.

ketch:
Oh, I didn't realize there was a physical meaning to negative pressures for the Tammann EOS.  If that's the case, I take back what I've said; I guess I don't know enough about it to give any recommendations.

I will read over the rest of the notebook again tomorrow.

ketch:
I've issued a PR with a few more changes.

By the way, this notebook is the only place where I've heard of this EOS referred to by the name "Tammann".  All the references I know call it the stiffened gas EOS.  What's the motivation for calling it the Tammann EOS instead?

maojrs:
Ok, I will check it and push the changes back up. 

Well there are other references that call it Tammann EOS, like: http://onlinelibrary.wiley.com/doi/10.1002/(SICI)1097-0363(19980915)28:3%3C395::AID-FLD718%3E3.0.CO;2-S/full

I really do not mind how we call it, so we can use the stiffened Gas EOS if it seems 
more suitable. Stiffened does sounds more descriptive, like a "stiff"(less compressible) gas, which 
makes sense so might be better. I only think it is important to point out the two names 
in the text so people know it is the same thing.

maojrs:
Done, tried to remove the extra metadata with the last commit, but it didn't work.

ketch:
Thanks for the reference!  I learned something new.  I think the presentation is fine, since you mention the other name of the EOS at the outset.

I've decided to just not worry about all those annoying metadata changes in the notebooks...

PR #72, opened 2017-05-07
Acoustics NB

maojrs:
Just made an update on the acoustics NB (not yet final). I tried to merge what was there before with some things I wrote before and some other new things. Hopefully it is not too messy. Some comments:

- Added simple intuitive derivation of the equations. Although this is available in many places, I thought acoustics is such a landmark example that it is worth to include.
- Added simple exact solution, before it was calling the clawpack solver (we want to be standalone from clawpack right?).
- Added exact solution for heterogeneous media/discontinuous coefficients (not sure what is the best way we should call it).
- Plots of particle trajectories work fine when I change in riemann_tools the line: "v_mid = 0.5*(q_old[i_vel,:] + q_new[i_vel,:])", for "v_mid = 0.5*(q_old[i_vel] + q_new[i_vel])". Not sure yet what is happening, I will take a look but maybe you have a fast answer.
- Still missing to add the phase plane plots, I am planning to do something similar (interactive app) to what I did with the euler_tammann notebook.
-Still missing an example or two for heterogenous case.

Comments, critics, what should be removed/added are all more than welcome.


ketch:
@maojrs What's the status of this?  I've been waiting before reviewing it, since the Travis build is failing -- looks like some of the modified code in the notebook throws an error.

maojrs:
Sorry, I was waiting on your comments before making more changes, and I didn't notice the Travis build failed. I will take a look into it.

maojrs:
Made a major update to this notebook. In order to get the particle trajectories to work, I had to do a small modification to riemann_tools. The particle trajectory plotting also seems a bit cumbersome, at least how I got it to work. It would be nice to have to call only one routine that does the whole plotting. The plotting of characteristics is not fully working because I need some modifications I made in randy's TammanNB branch to be able to pass several aux variables to the characteristics. Comments and suggestions are more than welcome.

rjleveque:
@maojrs:  Thanks for working on this!

I realized at some point that plotting particle trajectories makes no sense for acoustics: (p,u) represent some very small perturbation to the background state and the solution is invariant if you multiply by some constant (whereas `c` is fixed) so `u` should be essentially zero relative to `c` and all the particle trajectories should be almost exactly vertical.  The fact that they aren't is because we use O(1) values of `p` and `u` so these trajectories gotten by integrating `u` are meaningless.

I suggest removing these parts of the notebook, and also all the old stuff at the bottom.  I also suggested a few minor changes to text in a PR to your branch.

I'll merged #101 so you can fix the characteristics, but now I see this PR has some conflicts, not sure why...

maojrs:
@rjleveque thanks! That makes sense, I will remove the particle trajectory plotting then and the old stuff as well. I'll start working to get the characteristics right.

maojrs:
Ok, finished with the characteristic plotting. I had to make a tiny modification to the Tammann NB (latest version), so these new modifications don't break it. I included that modification here since it is affected directly by this branch. The modifications done to riemann_utils are quite general and should not break any previous functionality. These modifications basically allow the characteristic plots to include an interface (given by left and right aux variable(s) and maybe a contact discontinuity, depending on the case).

Made some other general modifications to the NB, removed the particle trajectory plotting and the old stuff, and added an additional example for the heterogeneous case.

One thing we should think about: should we keep the case with jump on rho and K here or move it to the "part III: Riemann problems in heterogeneous media"?

ketch:
There are already characteristics with an interface (different aux variables) in the variable-speed traffic notebook. Why did you need new code for this?

rjleveque:
Regarding the heterogeneous media, I do think we should move that part of the notebook to a new notebook in Part III, with a link to it from the current notebook.

maojrs:
@ketch if I am correct, the variable-speed traffic notebook only uses one aux variable (one left, on right). The heterogenous acoustics and the Tammann NB need two aux variables (two left, two right). The modifications I did was to accommodate these ones. In the case of the Tammann, I also need to pass the speed of the contact disconituity, so I also added that functionality. I was careful to add it in a way it doesn't break any of the previous functionality.

@rjleveque ok, I will start moving that part to a different notebook

ketch:
Doesn't the nonlinear elasticity use multiple aux variables and plot characteristics?

maojrs:
@ketch Yes, it does. I just figured it requires passing the aux variable in a very specific format to not yield an error. Anyway, rewrote the calls of Acoustics and Tammann to work with the original riemann_tools. Nonetheless, passing the contact discontinuity speed is still necessary for the Tammann notebook, so I left that modification in rieamann_tools.

Note I also made some changes to the notation for some function calls for acoustics and Tammann to be consistent with other examples.

rjleveque:
There was a merge conflict of meta data I created by merging #103, so I just fixed this.

This seems to work on all the notebooks I tested, so I'm going to merge it so we can move forward.

PR #71, opened 2017-05-06
Merged two PRs and added make_html.py

rjleveque:
I merged the PRs #68 and #69 into this one since they were starting to diverge and I wanted to test various changes from each of them together.  I suggest we close both of those if @ketch and @maojrs are happy with this one, and then merge this in and move on from there with new PRs.

I added a `make_html.py` to create html files from all the notebooks and put them in a separate directory `riemann_book_files` that can be moved to a website, such as: http://faculty.washington.edu/rjl/riemann_book/Index.html

This is far from perfect since much of the rendering is not working properly.  See for example [the Tammann](http://faculty.washington.edu/rjl/riemann_book/Euler_equations_TammannEOS.html) notebook, which shows many of the problems:

 - bib citations mess up the rest of the text beyond
 - equation numbers aren't showing up
 - javascript widget broken.

On the other hand the [Introduction](http://faculty.washington.edu/rjl/riemann_book/Introduction.html) looks pretty good except for the equation numbers.

PR #73, opened 2017-05-16
Make html test

ketch:
This is identical to #71 except that (since it said there were merge conflicts) I went ahead and merged master into it locally.  Strangely I didn't get any conflicts; I'm guessing that nbdime was able to resolve things better than ordinary diff could.


ketch:
Okay, the rasterized graphics are gone now.  See a working copy of the acoustics notebook here:

http://nbviewer.jupyter.org/gist/ketch/e167193a920f32308ab32535bc70ba56

ketch:
We have two different approaches to plotting stripes; one is in Euler_equations.ipynb and the other is in Shallow_tracer.ipynb.  We should probably pick one and move it into riemann_tools.  One issue with the version in Euler_equations.ipynb is that it rescales the axes for each frame.

I think we can merge the present PR before dealing with this issue.  This PR is already a monster...

Kitchen_sink_problem.ipynb
```diff
@@ -10,11 +10,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
```
ketch: Somehow this tag got deleted.  We need it.

Traffic_flow.ipynb
```diff
@@ -68,11 +67,37 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "![](./figures/LWR-Velocity.png)"
```
ketch: This figure was inadvertently duplicated.

ketch: I believe this was caused by a bad merge...not sure if nbdime is the culprit.

Traffic_flow.ipynb
```diff
@@ -68,11 +67,37 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
```
ketch: This cell also got duplicated...

Traffic_flow.ipynb
```diff
@@ -141,8 +166,8 @@
     "    plt.show()\n",
     "    \n",
     "interact(jam,\n",
-    "         rho_l=FloatSlider(min=0.,max=0.9,value=0.2,description=r'$\\rho_l$'),\n",
-    "         t=FloatSlider(min=0.,max=1.,value=0.2));"
+    "         rho_l=widgets.FloatSlider(min=0.1,max=0.9,value=0.2,description=r'$\\rho_l$'),\n",
```
ketch: This is reverting changes that I made in the bookform PR.  Was this intentional?  I suspect not.  I'm very worried now that much of my work will be lost unless we go over this PR with a fine tooth comb...

ketch:
It would be really helpful if we could avoid all the noise we've been getting in our commits.  A large portion of the "changes" in the present PR are just modifications to the "collapsed" and "deletable" metadata values.  I try to avoid committing changes to those (though I have messed up, and will do better), by always doing `git add -p` and then not committing the meaningless stuff.

The conflicts present in #71 are also due entirely (I think) to these meaningless metadata changes.

ketch:
This PR seems to clobber many changes I made to the Traffic flow notebook.  I haven't finished reviewing the rest of it yet...

Issue #75, opened 2017-05-18
Misc. problems


rjleveque:
Noticed while going through #74...

- [ ] Introduction.ipynb: double rarefaction at earliest t shows artifacts at discontinuity,
- [ ] Introduction.ipynb: The interacts have `t = 0.001` to avoid problems at `t=0` and these are not reset to `t=0.2` by `utils.snapshot_widget`. 
- [ ] Intro and elsewhere: `utils.snapshot_widget` creates some plots where it's not clear what time the solution is shown at, do we want to include t in titles?
- [ ] Traffic_flow.ipynb: Traffic jam shock leaves the domain: expand xlim?
- [ ] Shallow_water.ipynb: Why do cells show tags (only in this notebook)?
- [ ] Shallow_tracer.ipynb: Redo with the new plotting of stripes rather than using PyClaw? Needs more description too.

ketch:
>Shallow_water.ipynb: Why do cells show tags (only in this notebook)?

You can turn the visibility of tags on and off through the menu: View -> Cell Toolbar -> Tags

I probably saved it with this on by accident -- sorry!

ketch:
>Shallow_tracer.ipynb: Redo with the new plotting of stripes rather than using PyClaw? Needs more description too.

That chapter is just a placeholder; it's not in the PDF yet.

PR #76, opened 2017-05-22
Improve presentation of Introduction in PDF.

ketch:
- Resolve issues with t=0 (see #75)
- Make all code fit on page by shortening lines and reducing code font size
- Avoid using top-level header (#) since that gets converted to a new chapter

A PDF showing how things look with these changes is here:

https://www.dropbox.com/s/9huu3ig543ut528/riemann.pdf?dl=0


ketch:
I also changed the references to other chapters to make more sense when converted to PDF.  Since markdown links like [Shallow_water.ipynb](Shallow_water.ipynb) are converted to something like "Chapter 5", I removed the phrase "the notebook" from before them.  Alternatively we could write [the notebook Shallow_water.ipynb](Shallow_water.ipynb); the latex results would be the same (whatever is between the square brackets just gets removed).

rjleveque:
Looks good!  I think it's fine to remove "the notebook" as you have done.

Issue #79, opened 2017-05-22
Electronic material that needs to be adapted for the print book


ketch:
There are a few things that still read funny in the PDF and need to be modified to work there:

- [ ] At the end of the introduction, the interactive phase planes for acoustics and shallow water.  Can we host these HTML files somewhere (Github doesn't seem to work for that) and put a hyperlink in?

rjleveque:
I suggest we create a gh-pages branch for showing the rendered notebooks.

I did this on my fork, visible at http://rjleveque.github.io/riemann_book/

If @ketch and @maojrs like it, I can push the branch to this repository.

ketch:
I really like this!  Please do push it to the main repo.

rjleveque:
Done.  Now visible at: http://www.clawpack.org/riemann_book/

There are still some glitches in the html files I'll work on some more.

PR #82, opened 2017-05-31
Move phase plane from intro

rjleveque:
Now point to the separate html file so this looks better in the pdf file, addresses #79.


ketch:
Thanks!  It would be nice if the PDF hyperlink also worked, but we would want a full URL link for that.  Maybe we can point to the gh-pages copy?

rjleveque:
@ketch: The link to the phase plane should work once #83 is merged.

rjleveque:
I haven't tested this in the pdf yet since `bookbook` stopped working for me...

ketch:
I'm not sure why this still shows as open after I merged #83.  There must be some additional commits here?

rjleveque:
Yes, this is distinct from #83 -- this modifies two notebooks to point to the webpage.

rjleveque:
Oops, I take back my previous comment.  What's different is that this PR merges into `master` while #83 merged into `gh-pages`.  

Going forward we need to remember to merge `master` into `gh-pages` before updating the html files for webpages.

PR #90, opened 2017-06-06
Revise later chapters (6+)

ketch:
I'm currently making a pass through some of the later book chapters, especially improving things for the PDF.  This isn't ready to merge yet.

PR #91, opened 2017-06-07
[WIP] Revise early chapters

rjleveque:
I'm working on revising some of the earlier chapters for the pdf, to complement #90.

Not ready to merge yet.


rjleveque:
@ketch: I shortened some lines by indenting continuations of function call argument lists differently, with less white space before the continued lines so that some calls would fit and/or would fit into fewer lines.  Particularly in `Shallow_water` and `Euler_approximate_solvers`.

If this and other changes seem ok, this could be merged.

ketch:
I'm a very strong believer in lining up continuations of function arguments with the parentheses for the function, and this breaks that.  But I don't know what the optimal solution is, and I suspect we'll revisit this before the book is finished, so I'll go ahead and merge this.

PR #92, opened 2017-06-07
Refactor Advection.ipynb

rjleveque:
A proposed change to this notebook: I moved the code from the bottom of `Advection.ipynb` to `exact_solvers/advection_figures.py` to clean this up a bit.

I also adjusted the last plot so it is clearer (I think) that the discontinuity is moving along the characteristic from `x = 0`.


ketch:
This looks very nice.  I guess you are proposing that we go the route of keeping plotting code out of the notebooks as much as possible?  Or is the idea just to spare the reader from looking at code in this early chapter?

ketch:
Oh, you committed this with the output included -- was that intentional?

rjleveque:
Oops, output now removed.

The original version of this notebook had png files of the figures and then at the end a few cells labeled "Code used to generate figures".  I thought this was a cleaner solution to moving the code out of the way, since code for these figures isn't very interesting to most readers but from the import  statement it will be clear where to find it if they want to use it for something.

ketch:
Yes, I parked the code there originally as a reminder that we needed to decide where to put it.  This looks fine.

PR #94, opened 2017-06-07
Build PDF in a subdirectory with make_pdf.py.

### LINKED ISSUE #84: Create pdf in build_pdf directory?
@ketch: I suggest we 

- consolidate `make_chapters.py` and `make_book.sh` into `make_pdf.py` similar to `make_html.py` so it's less confusing where things are done and where the specify the desired chapters.

- Build the pdf file in a new directory `build_pdf` so that all the files generated `0*.ipynb`, `riemann.aux`, etc. don't end up cluttering up the main directory.  Then move the `pdf` file out of this directory so it can be safely deleted to clean up.  This would be similar to what's done in #83 for html files.

ketch:
Eliminates `make_book.sh` and `make_chapters.py`.

Resolves #84.


rjleveque:
Works for me, thanks!

PR #95, opened 2017-06-11
remove obsolete make_html.py from master 

rjleveque:
Correct version on gh-pages branch.  That is the branch where it is used.


rjleveque:
I'm merging this so I can fix up the gh-pages branch.

PR #96, opened 2017-06-11
fix </a> at end of data-cite bibtex references

rjleveque:
Bibtex references using `data-cite=...` have to end with `</a></cite>`, not `<a></cite>`.

These notebooks looked fine as notebooks but rendered wrong as html.

### DIFF
--- a/Acoustics.ipynb
+++ b/Acoustics.ipynb
@@ -7,13 +7,15 @@
     "\n",
     "# Riemann problem for acoustics\n",
     "\n",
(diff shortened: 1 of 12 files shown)


rjleveque:
@ketch: Remind me how we create the `riemann.html` file that the citations point too?  My version of that file dates from February and is missing most references.

Do we run `bibtex2html` on `riemann.bib`?

ketch:
Thanks for catching that!  I must have written the first one wrong and then we copy/pasted it.

Instructions here: https://github.com/clawpack/riemann_book/wiki/Citations#how-to-generate-the-html-bibliography

PR #98, opened 2017-06-12
Add html subdirectory

rjleveque:
Add all the html files to display at
http://www.clawpack.org/riemann_book/index.html
and fix this index so it points to `html` subdirectory rather than `riemann_book_files`.

Also merged in various changes to `master`.


ketch:
Ummmm....44 commits and 323,000 lines added?  You've been working hard!

Is there any possibility of streamlining this?  The diff seems to be mostly noise (meaningless metadata changes) and duplications of other PRs.  I can't find the actual content.

rjleveque:
@ketch: Note that this is a PR on the `gh-pages` branch, which was way behind `master`, so most of these are commits that were already merged into `master` in previous PRs but need to carry over to `gh-pages` in order to have the up-to-date notebooks to create the html version.

In the future perhaps I should just merge master and push such changes directly without a PR when updating the html version in `gh-pages`, and only do a PR when there's some change to `make_html.py` or something else that others might want to review?

In the case of this PR, the main changes are to `make_html.py`, `index.html`, and moving all the html output to a subdirectory `html`.  I also committed all those html files for the first time so that they will show up on the webpage.

Sorry for the confusion!

ketch:
Ah, I completely missed that this is a PR to gh-pages rather than to master!  Please just disregard my comments; this is fine.

PR #99, opened 2017-06-12
Nonconvex scalar

rjleveque:
Improvements to Nonconvex_scalar notebook (also renamed).

Now uses widgets like other notebooks and converts to JSAnimation in the html version.

Also added convex hull plots and refactored the functions.

Might want to move the plotting functions to a Python module?  As it is, there's a lot to skip over before getting to the figures.  But in this instance I think the reader might be interested in how the various plots are created.

### DIFF
--- a/Index.ipynb
+++ b/Index.ipynb
@@ -36,7 +36,7 @@
     "## Part IV: Source terms\n",
     " \n",
     "## Part V: Non-classical problems\n",
-    "  1. [Nonconvex flux for a scalar problem](Nonconvex_Scalar_Osher_Solution.ipynb)\n",
+    "  1. [Nonconvex flux for a scalar problem](Nonconvex_scalar.ipynb)\n",
     "  2. [Pressureless flow](Pressureless_flow.ipynb)\n",
     " \n",
     "## Part VI: Multidimensional systems\n",
@@ -44,15 +44,6 @@
     "  2. [Elasticity](http://nbviewer.jupyter.org/github/maojrs/ipynotebooks/blob/master/elasticity_riemann.ipynb)\n",
     "  3. [The Kitchen Sink: shallow water in cylindrical coordinates](Kitchen_sink_problem.ipynb)"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/Nonconvex_scalar.ipynb
+++ b/Nonconvex_scalar.ipynb
@@ -32,7 +32,9 @@
    "source": [
     "## Define a function returning the  Osher solution\n",
     "\n",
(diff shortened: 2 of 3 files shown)


rjleveque:
@ketch: I haven't tested this one in the pdf version since `bookbook` quit working again for some reason, I get error messages from `inkscape`, first some warnings that figures like 
```
** (inkscape-bin:54814): WARNING **: Can't open file: ./tmpy_8zk45w/figure.svg (doesn't exist)
```
and then
```
    raise TypeError("Inkscape svg to pdf conversion failed")
```
This doesn't seem to be related to adding this particular notebook, so no idea why it stopped working.

Anyway, if the pdf version looks ok we might want to include this notebook.

rjleveque:
I got `bookbook` working again (just by opening a new shell, don't know what was messed up).

The pdf for `Nonconvex_scalar.py` looks ok to me, other than perhaps too much code?

ketch:
I just read through this; it's great.  I have a few comments:

- If you want to leave code (or anything else) in the notebook but don't want it in the PDF, just add the tag "hide" to that cell (dropdown menu view -> cell toolbar -> tags).
- In the last cell in the section "Traffic flow", I was initially confused by the phrase "we see initial data" since the initial data isn't plotted there (though it can be inferred).
- In some of the plots (like the flux and derivative) things are a bit crowded.  I think if you add `plt.tight_layout()` to the end of the cell, it will look nicer.
- The use of red and blue in the plots here seems like it contradicts how we use them in the rest of the book.  Would it make sense to plot the multi-valued solution as a black dashed line (which we use for entropy-violating things) and the correct weak solution as a solid black line?  For the flux and convex hull, perhaps we could just use some other colors, like I did in the variable-speed traffic chapter (though I don't know if I picked the best colors there).

rjleveque:
Thanks @ketch, good ideas.  I'm not sure black solid and black dashed is a good combination, but I'll play around with the colors some more and the other things and then update this PR.

ketch:
And yes, I agree that we should add this to the PDF.  We should probably also remove the chapter on Euler approximate solvers for now.  After that I think we are ready to submit.

rjleveque:
Some improvements:

- I hid some code cells and added a comment about the functions they define, since otherwise a reader of the pdf might be confused since it seems like the `osher_solution` function isn't actually used and not know where these other functions came from.  Should we perhaps indicate in the pdf somehow where hidden cells occur?

- I modified the traffic flow description.

- `plt.tight_layout()` didn't seem to change things in the notebook, does it only affect the pdf?  I increased the width of this figure, which did help.

- I changed from blue/red to solid/dashed and it looks better than I expected.

Eventually I'll add more general discussion at the beginning of this notebook about the issues of nonconvex  fluxes, but I think this is ok for now.

rjleveque:
Apparently this wasn't passing in Python3, but the error seems to be from Acoustics.ipynb something to investigate before merging #72.

ketch:
BTW, for me `tight_layout()` changes the spacing in the notebook.  Maybe it's an issue with screen resolution and browser window size.

PR #101, opened 2017-06-21
Fixed some problems in Euler_equations_TammannEOS.ipynb 

rjleveque:
I also raised some issues in the form of markdown cells with red text that I added, mostly about the last few examples. I don't think they are explained properly and I'm not sure what exactly was intended regarding discussing shocks passing through an interface.

Also it seems like the left-going shock should be a rarefaction in some cases?

### DIFF
--- a/Euler_equations_TammannEOS.ipynb
+++ b/Euler_equations_TammannEOS.ipynb
@@ -411,7 +411,7 @@
    },
    "source": [
     "### Problem 1: Sod shock tube in water\n",
-    "We first consider the classic Sod shock tube with higher density and pressure on the left state than on the right with both sides initially at rest. The main difference with the classic problem is that in this case the shock tube is filled with water and not with air. The propagation of waves in water can be accurately modeled using $\\gamma = 7.15$, $p_{\\infty} = 3\\times 10^8$ and $\\rho\\approx 1000 kg/m^3$; see <cite data-cite=\"del2016computational\"><a href=\"riemann.html#del2016computational\">(del Razo 2016)</a></cite>. A value of $p_{atm}=101325 Pa$ is used for Atmospheric pressure."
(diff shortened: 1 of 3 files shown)


maojrs:
Randy, I committed directly into your branch by mistake. I thought it was gonna push it directly into my fork, so sorry for that. Anyway, I addressed some of the issues you mentioned in the examples. The example with the left rarefaction is indeed a rarefaction. However, the problem was the tail/head velocities were really close to each other. I changed the parameters to make the rarefaction more visible. I also changed the discussion of the other examples to address your comments. Hopefully, it is better now. If you think still think some things should still be changed, I am happy to do the modifications.

rjleveque:
@maojrs: Very strange that you could push to my fork since you're not listed as a collaborator, I don't know how that happened!

I made some more changes but some things seem wrong:

Problem 1: I added a note that we use a large jump in p to illustrate the rarefaction wave. However, when you turn on plotting of the characteristics, they look wrong.  Presumably the 1-characteristics should be spreading across the rarefaction wave and the 2-characteristics should be parallel to the contact discontinuity.

Problem 2: I tried to better describe the setup for calculating transmission and reflection at an interface.  In this example the 3-characteristics look wrong, they don't impinge on the shock.

Problem 3: All the characteristics look wrong, and why does the left-going wave have a jump in pressure but not in density or velocity?

Strong rarefactions: The 1- and 3- characteristics seem wrong.  Would a less extreme set of data avoid the errors `RuntimeWarning: invalid value encountered in power`?

maojrs:
@rjleveque yes, I thought I wouldn't be able to push to your fork as well. Anyway, I will be more careful that I am actually pushing to my fork.

It is likely there are some issues with the characteristics. When I wrote the code to generate the characteristics, I was not able to test them since the characteristic plotting was not working then. I will review them.

Regarding Problem 3: I think there is indeed a jump of velocity and density; it is just too small to appreciate in the plots, maybe a different set of parameters might make it more evident. I will give it a try.

maojrs:
@rjleveque I just made a pull request to your branch that fixes the characteristics plotting. Hopefully, there are no more issues. Please let me know if you find anything else. The "RuntimeWarning: invalid value encountered in power" is a bit unpredictable; most of the times it doesn't appear, so I am not really sure if it might be a jupyter bug or something else. Regardless, everything seems to be ok even with the error, though I will investigate further.

rjleveque:
@maojrs Thanks for fixing the characteristics, looks great now!

I added a bit more discussion of the pressure p as deviation from the ambient pressure (so negative values make more sense) and a new cell at the bottom that prints out the values of p and `p+p_inf` that are discussed in the last paragraph.  Does this seem like a good idea?

I also added a bit of discussion about why `gamma` and `p_inf` are constant across the nonlinear waves.

rjleveque:
I'm merging this so @maojrs can use it in making characteristic work in Acoustics.

PR #105, opened 2017-07-05
Fix permissions + make docker image much smaller

yuvipanda:
Docker stacks based Dockerfile was over 6g, making it really large.
This takes quite a long time to build on user's computers. The
newer one is simpler & is only 1g, much faster to build.

Also fixed issue with permissions in home directory, & ipywidgets
not working.


rjleveque:
Thanks, notebooks seem to work fine on docker container now.

PR #107, opened 2017-07-06
Shallow water plots

rjleveque:


rjleveque:
Merging for now but this notebook needs more work -- the new plotting doesn't work yet for the nonphysical rarefaction case.

Might also want to move more  of the plotting code out of the notebook.

PR #108, opened 2017-07-07
Fixes for html

rjleveque:
Fix some things that break in the html version.

For future reference:
 - Make sure to have `</a>` at end of citations,
 - Leave blank line before and after displayed equations in markdown
 - make sure `from ipywidgets import interact` is on a separate line. Use another line to import other things from `ipywidgets` since this line is replaced when converting to html or pdf.

### DIFF
--- a/Acoustics.ipynb
+++ b/Acoustics.ipynb
@@ -2,23 +2,17 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


ketch:
Leaving a blank line after displayed equations will "break" the pdf because it will put a paragraph break there, which is often inappropriate. Markkdown recognizes two spaces at the end of a line as a line break; does that help?

rjleveque:
Yes, I've noticed we have new paragraphs starting after most equations in the pdf.  I'll experiment with the two-space option.  Thanks!

PR #115, opened 2018-02-13
Burgers approximate solver chapter

ketch:
This PR adds a chapter on scalar approximate solvers using Burgers' equation as an example.  I also added a short section on notation in approximate_solvers.ipynb.

The writing is quite rough in some places, but I would appreciate feedback on the draft at this point.  The chapter is very short; you can read through it in a few minutes.  I haven't added any computational examples; I'm not sure whether there is anything useful to add in that regard.

If either of you want to go ahead and try to smooth out the rough sections, that would of course be welcome too.

It would be nice to have the theoretical chapter on Burgers' equation, in order to refer to it.  My notes from our call in October say that Mauricio was going to work on that -- any progress?

I will be reworking the Euler approximate solvers chapter next, which I imagine will be our most thorough chapter in this section of the book.


rjleveque:
Thanks for getting back to this @ketch, looks good.

I just made a couple suggested changes on this PR:
https://github.com/ketch/riemann_book/pull/5

ketch:
Another thought on subscripts vs. superscripts: we use subscripts elsewhere in the book for characteristic field indices (on eigenvalues and eigenvectors of the flux Jacobian).  Because of this, I'm now leaning strongly toward using subscripts on the waves and speeds.  Let me know what you think.

maojrs:
Also thank for getting back to this @ketch. Yes, we mentioned indeed in the last meeting I will work on the Burgers' notebook. I have not yet worked on it, but since we are back on it, I will start this week. 

Regarding the superscripts vs subscripts, I think it is better to keep consistency with the literature. I think people might be referring back to it, and it could be really confusing if it changes from super to subscript from one place to another. I think as long as we mention clearly that we are talking about 1, 2 or more waves in the text, it should be fine. Also the first superscript that appears is always one, so one can tell immediately they are not exponents. Nonetheless, I am open to change them if we agree it is better.

PR #116, opened 2018-02-14
Initial draft of chapter on approximate solvers for shallow water and Euler

ketch:
I decided to go ahead and draft the SW chapter first; Euler will be next.  I think this is fleshed out enough that I'd like feedback on it.  Some notes:

- There is a lot of code in the notebook.  Some of the boilerplate code that is repeated for each example can, of course, be put in a function and hidden away.  I think we should leave the Riemann solver implementations in the notebook; alternatively we could put each in a file and use the magic command to insert the contents of the file in a cell so readers can easily peruse it.
- There are probably too many examples, but I figured I'd include them all and then we could pare it down after discussing what is most important.
- I wrote numerical solvers specifically for this chapter even though similar Python solvers exist in clawpack/riemann.  This is partly because the way we're dealing with entropy fixes in the book is not a practical way to do it in a numerical code, and partly because the solvers in Riemann are vectorized and I think it's a lot more natural to include pointwise solvers in the book.

A note about useful Jupyter notebook extensions:  To more easily navigate around the notebook when there is so much code, I've found it very helpful to use the "table of contents (2)" and "collapsible headings" notebook extensions.  I'm trying to use lots of subsections in order to make this even more useful.  I've also found the scratchpad extension to be nice, in general.


ketch:
Also, my superscripts and subscripts are a mess because I changed my mind about which was better while writing this chapter.  But I didn't go back to make it consistent because I wanted to wait for your feedback on that question.  In this chapter it actually seems nicer to use superscripts for the characteristic index, because when discussing the entropy fix we also want to use a subscript for left/right.

ketch:
I just added a new final section to this chapter, in which the Roe averages for SW are derived purely from the requirement that the solver be conservative.  @rjleveque I have a couple of questions:

1.  I worked this out because I remember hearing you state the connection between conservation and Roe averaging, but I wasn't sure (until I worked it out) that the Roe average was the only average that gives conservation.
2.  Technically it turns out that there is a second "Roe average", that is a different velocity but is equivalent to the usual one when used in the solver (i.e., the resulting waves and states are the same).  This is of course just a curiosity, but I hadn't heard of it before and wondered whether it was known?

rjleveque:
@ketch: I knew that if you use an average other than the Roe average then in general the wave-propagation form won't be conservative, but I never tried deriving the Roe average by this means, and wasn't aware of the second possible choice.  This seems like a simpler derivation that the parameter vector approach introduced by Roe and that I followed in FVMHP.  I wonder if this approach makes it easier to derive Roe averages for other, more complicated, systems.  Have you tried it for Euler?

ketch:
@rjleveque Thanks for taking a look.  This seems rather interesting now.  It does work for the Euler equations -- and you get two possibilities for the "average" there, too.  I'll add an Euler notebook with the derivation soon.

I also think it is a simpler derivation, but what I especially like is that everything follows from the requirement that $f'(\hat{q}) \delta q = f(q_r)-f(q_l)$ (sorry that this isn't typeset on Github).  There is no magical step of guessing a nice parameterization.  It's also very natural; what we're imposing is almost just requirement (iii) from p. 359 of Roe's original paper.  The only difference is that we additionally assume that $\hat{A}$ is just $f'(q)$ evaluated at some state $\hat{q}$.

It seems like it would be very surprising if nobody had done this before.  But you of all people would probably know about it if they had.  I will see if it works for more complicated systems; do you have any suggestions for one to start with?

ketch:
By the way, it was just today (while working this out for Euler) that I realized all I was doing was imposing $f'(\hat{q}) \delta q = f(q_r)-f(q_l)$.  Using this directly avoids a bunch of (silly, in retrospect) algebraic manipulation; I've updated the derivation in this PR and now it's extremely simple.

ketch:
Oh, and for Euler there are also two choices of Roe "average"; they look very similar to the two for shallow water.

ketch:
After a bit of looking, I found that people have worked this approach out before for the Euler equations.  This is mentioned for instance in Section 3.1 of this paper:

https://doi.org/10.1006/jcph.1997.5838

which in fact says that "several authors" have noticed this.  What seems to be the best extension to more complicate equations of state is in this paper:

https://doi.org/10.1006/jcph.2001.6915

They impose the same condition I'm using and find that for general equations of state you get the same Roe averages for u and H.  You also need a particular average of rho (unlike in the polytropic ideal gas case, where the average of rho does not come into play); this average is determined by an equation that depends on the particular equation of state.

ketch:
Update: This PR also includes two Euler notebooks now:

 - One introducing a Roe solver and an HLL solver, and showing what they give for a few Riemann problems.  This follows the same outline as the shallow water chapter on approximate solvers.  It still needs work and is missing some things (including discussion and implementation of an entropy fix).
- A second chapter showing results of a shocktube problem obtained by running PyClaw with the Roe and HLL solvers.  This may be outside the scope of the book; I'm open to suggestions in this regard.

ketch:
In the last commit I've addressed almost all the comments of @rjleveque .  There are probably too many / too large legends now, but we can keep refining that.

The one outstanding comment is about whether to plot u or hu.  In the earlier notebooks on SW and Euler we focused on primitive variables.  If you agree, I can change the approximate notebooks to do the same (plotting u rather than hu).

rjleveque:
We should discuss u vs. hu.  Each version is useful for certain things.  The phase plane plots look very different since u blows up while hu goes to zero as h --> 0.  

I've always used hu in the past, mostly because [h, hu] are the variables in which the eigenvectors are naturally defined, and these vectors are tangent to the integral curves in the h-hu plane, so this version fits better when describing the differences between linear and nonlinear Riemann problems, I think.

On the other hand, for problems with dry states appearing in the middle, the h-u plane shows what's going on much better.

PR #117, opened 2018-02-18
Traffic with an on-ramp

ketch:
This PR adds a notebook for the LWR model with an on-ramp at x=0.  This is intended to be the first chapter (other than perhaps introductory material) in Part IV of the book, which is focused on source terms.

There are several examples and a Riemann solver that I believe is correct for all inputs.  I used the f-wave solver from chapter 17 of FVMHP to check my work.  Eventually I found some situations for which that solver gives an incorrect answer; I haven't tried to write an approximate solver that is better.

The text is quite rough in places, with some repetition.  I haven't tried to fully explain every possible type of Riemann solution -- I'm not sure whether that would be worthwhile.

Feedback is welcome.


ketch:
@rjleveque I'm copying some of your comments here for continued discussion.  I believe I've dealt with the rest in my latest commit.

> It seems to me there is a fundamental difference between the conservation law with $f(\rho,x)$ for a variable speed limit, in which the total mass of cars is conserved, and the balance law that has a conservative flux on the left and a source term on the right. In the latter case the mass of cars is not conserved.

> It's true both have a delta function source on the right but it's not true that an on-ramp exactly corresponds to a decrease in speed limit. The on-ramp introduces new mass and does not really change the speed limit on the right since it's the same flux function, it just increases the density and hence the speed decreases due to driver behavior.


The text in question was intended as a placeholder to remind me to discuss this connection, rather than a final draft.  I do think it's worthwhile to say something about this connection; if you would like to rewrite this part, it would be welcome.  I did have in mind that there might be a precise correspondence if one also stretched the x-axis appropriately -- by different amounts for x>0 and x<0.  I haven't worked it out, though.

> In the next plot, the characteristics seem too dense and only extend up to t=0.4.

This is a feature of our current code.  All the characteristics from the region we plot impinge on the shock by t=0.4.  If this is undesirable, we could adjust the code to plot characteristics over a wider region.

> In the plot above, why do cars accelerate instantaneously at $x=0$? Shouldn't they accelerate slowly through a rarefaction?

Several things in this notebook are counterinuitive, but I believe they're correct.  In this case, it's evident that there cannot be a right-going rarefaction because the characteristics for x>0 are going to the left.

>should we also plot new vehicles entering at $x=0$ at various times?

Yes, this is something I'd like to add.

PR #119, opened 2018-02-20
Revise introduction.

ketch:
Add more historical notes at the beginning of the introduction.
Minor editorial changes to the text.
Examples of the Riemann solution as description of large-scale behavior.

Note on the last point: I wanted to add this as further motivation for studying the Riemann problem, after Constantine Dafermos mentioned it during a talk here at KAUST.  I'm not aware of a formal statement or theorem describing the tendency discussed in the notebook; if either of you are, please let me know.

The plots of the "macroscopic behavior" examples don't quite match the other plots, and are rendered as JSanimations currently.  I can fix this if/after we agree that this section belongs in the introduction.

### DIFF
--- a/Acoustics.ipynb
+++ b/Acoustics.ipynb
@@ -4,9 +4,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "\n",
-    "# Acoustics equations\n",
-    "\n",
+    "# Acoustics equations"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "This notebook illustrates some features of the exact solution to the 1-dimensional constant coefficient acoustics equations, as described in more detail in Chapter 3 of <cite data-cite=\"fvmhp\"><a href=\"riemann.html#fvmhp\">(LeVeque 2002)</a></cite>. Since acoustic equations are a landmark example of linear wave propagation, we will begin with a brief derivation of the equations. This will also help having a self contained document. "
    ]
   },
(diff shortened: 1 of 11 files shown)


ketch:
Update: the latest commit in this PR modifies the `snapshot_widgets` function so that if an interact has an argument `t` with value zero, then plots are produced for both t=0 and t=0.2.  This makes it so that the reader can see both the initial data and the Riemann solution.  Here's an example of the output:

screen shot 2018-02-20 at 4 31 17 pm

If one sets t to a non-zero value, then only the Riemann solution is plotted (not the initial data).

ketch:
Update: the latest commit adds a custom CSS file to modify the look of the notebook.  I made the following changes:

1. Use fonts similar to those of SIAM books.  Garamond for the body (this is an exact match) and Helvetica for the headings (this doesn't quite match but was my best guess).
2. Limit the width of rendered text in markdown cells to a maximum of 700 pixels.  I think this helps with readability in a wide browser window.
3. Add a dark grey background on the far outside margins (to suggest the feel of a book) and a faint off-white background around code cells (just to distinguish them from the text and provide more points of reference for the eye).

One thing that's a bit weird is that the code and plots are allowed to be wider than the main text.  I'm not sure if this is a good idea; it can easily be changed (to make the text wider or the code/plot cells narrower).

Try it out and let me know what you think.  Here's a screenshot:
screen shot 2018-02-20 at 10 14 07 pm

ketch:
I forgot to mention -- to apply the styling, you have to run the second code cell, which loads a CSS file.  I need to find a cleaner way to do this, so that the notebook loads with the correct appearance from the start.

ketch:
Further update: I discovered today that JSAnimation was merged into matplotlib version 2.1, last fall.  In the last commit I've switched over to using this.  As far as I can tell, the fixed the issue where you would sometimes get a javascript error instead of the animation (hooray!)  You will need to update to matplotlib 2.1+ for this to work.  The modified version of animation_tools.py should be backported to visclaw (and we can delete the JSAnimation directory there).

There is also the option to convert animations to html5 video.  The main difference in practice is that you only get a play button:

screen shot 2018-02-21 at 9 01 03 am

I think I prefer the JSAnimation output (and I have stuck with that), since it allows the reader to more easily step to any particular frame.

rjleveque:
I like the revisions to the Intro, and I think I like the new CSS style.  It's a bit disconcerting that the code cells and plots are so much wider than the text.  I like the wider text from the old style because it means more text fits in a screenfull and not so much scrolling is required, which can be annoying when trying to look at a figure and also read the text that describes what you're seeing.  I suppose the text could be made a bit wider?

ketch:
Thanks.  I found the limited text width disconcerting too, so I've now just let the width go as wide as the browser.  Readers can always get a narrower view by narrowing the browser window size.

ketch:
I'll go ahead and merge this for now.  We can continue adjusting the style as needed in the future.

PR #120, opened 2018-02-21
Plot characteristics for shocks and rarefactions in nonconvex notebook.

ketch:
This is implemented in a fairly crude way for now -- I just take the solution q(xi) and try to detect where it is constant, continuous, or discontinuous.  Then shocks are plotted at points of discontinuity and rarefaction fans over continuously-varying regions.  It's somewhat sensitive to how many points one chooses to sample, since it's trying to determine smoothness based on differences.

Here's a screenshot:

screen shot 2018-02-21 at 8 54 31 pm


ketch:
@rjleveque regarding your comments on this update: I agree, there are several issues with the plots.  I just wanted to check whether you thought this was the right general direction before I spent time fine-tuning things.

PR #124, opened 2018-02-25
Interact improvements

### LINKED ISSUE #122: Error pops up in plot_riemann for Traffic
Error comes up when plotting Riemann problem. For instance, when plotting the last example in traffic_flow notebook, if densities are chosen to zero, an error pops up. I think this has to do with how the axes-ranges are chosen. I also think it might be more educational if the x-axis range doesn't change when modifying the Riemann's problem initial condition. This way one can observe how the shock/rarefaction changes as function of the I.C. This concerns also the Burgers' equation notebook that I am working on.

ketch:
This PR fixes some situations in which plot x- or y-ranges were set to have the same min and max values (which caused matplotlib to print a warning) in riemann_tools.py.

It also adds a keyword argument xmax to plot_riemann(), allowing the user to specify a range for the x-axis rather than having it scale automatically.  This seems generally better for interacts.

Finally, I added an improved interact widget to the first traffic notebook, at the end.  The x-axis range control is included, and different sliders are put in separate columns.  I left the code for this in the notebook for the moment for easy inspection, though we'll want to move it to a file later.

Thanks to @maojrs for pointing out these issues and for showing me (via his interactive pplanes code) how to lay out better widgets.

I think this resolves #122.

### DIFF
--- a/Traffic_flow.ipynb
+++ b/Traffic_flow.ipynb
@@ -252,7 +252,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": false,
     "tags": [
      "hide"
     ]
@@ -323,7 +322,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": false,
     "tags": [
      "hide"
     ]
@@ -369,7 +367,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": false,
     "tags": [
      "hide"
     ]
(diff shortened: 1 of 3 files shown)


maojrs:
Thanks @ketch for addressing this. It seems everything is working ok, except for the vehicle trajectories in the interactive app (traffic notebook). If one scales the x-axis the vehicle trajectories plotting doesn't seem to update the range in which the trajectories are actually plotted.

ketch:
@maojrs Thanks for noticing that.  I fixed it (just needed to pass the xrange keyword into the trajectory-plotting function).  I'll go ahead and merge this since you're already wanting to use it in another PR.

PR #125, opened 2018-02-28
Burgers equation NB

maojrs:
This pull-request adds the Burgers' equation notebook. The notebook is a first draft, though it is essentially complete. However, I would like to see if you have any suggestions or comments. I have a couple of additional ideas regarding this:

- Maybe Burgers eq. notebook should go before the traffic notebook in the index. The traffic eq. is a more general version of essentially the same equation. However, I understand the traffic eq. provides a great example physically, so I am not entirely sure about this.

- I thought of writing a solver for a three-state initial condition for the Burgers' eq. I think this could be really cool because it is the simplest scenario where we can observe rarefactions interacting with shockwaves and so on. We could also plot the characteristic for these examples, and it would really be enhanced by the interactivity. However, it cannot use the standard Riemann tools, so it would require some additional work. The question is if it's worth to write a general plotter for a three-state IC system, or is it better to only do it for the Burgers' eq.

Note this notebook benefits from the pull request Interact improvements #124


ketch:
This looks like a good start and has a lot of nice figures and discussion.  I think it works nicely as a standalone discussion; as you suggest, it needs to be adapted to fit into the book.

I would like to keep the ordering we had previously agreed on (traffic before Burgers), for two reasons:

1. I like starting with an example that people have some experience with and intuition for.  It allows many of the concepts to be introduced in a very natural way.

2. The traffic chapter introduces several important concepts for the first time.  It would need significant revision if it were placed after Burgers.  It has already gone through several revisions, and while it can still be improved, I'd rather not rewrite it again to that degree.

I think with some minor changes to the text (to cue the reader that these concepts are not totally new), the illustrations and discussion you have included (like wave breaking and the equal-area rule) can nicely complement what was already introduced in the traffic chapter.

One thing I think we had planned for the Burgers chapter was to show multiple weak solutions of a single Riemann problem.  This serves as a foreshadowing of the need for an entropy fix, which we also introduce in the context of Burgers' equation.

Showing interactions between waves emanating from two different Riemann problems might be nice.  One option would be to infer the waves from a numerical (or iterative) solution, rather than writing an extended exact solver.  I did something sort of like that in #120; we could probably adapt it to do what you're thinking of.

ketch:
Two other comments, after re-reading:

1.  You might mention specifically that Burgers' equation models momentum transport in a fluid of uniform density and pressure.
2.  The explanation of the formula for a rarefaction is very nice.  The text seems to imply that the first rarefaction shown is not a centered rarefaction.  But as I understand the term, all rarefactions appearing in the Riemann solution are centered rarefaction waves.  "Centered" doesn't refer to it being symmetric or transonic, but to the fact that the characteristics emanate from x=0.

maojrs:
@ketch thanks for taking a look at it. Regarding your comments:
-Lets keep the same order then, and I'll rewrite the notebook so it fits better into the book. I will also take a look of what can I do for the multiple weak solutions.
-Thanks for suggesting #120. I'll see what you did there and see if I can adapt it to show the three-state IC solution
-I will add the momentum transport comment
-I will correct the centered rarefaction, maybe call it symmetric expansion/rarefaction.

PR #130, opened 2018-03-13
Revise parts 1 and 2

### LINKED ISSUE #127: Use of red and blue colors in shallow water plots
Currently we use red and blue to distinguish water that started to the left/right of x=0.  We also use these two colors to distinguish shocks/Hugoniot loci versus rarefactions/integral curves.  I wonder if it would be better to change one of these to prevent the reader from thinking that they are somehow related.

For instance, we could just use blue and green in the shallow water plots.

ketch:
This is a major revision of most of the chapters in the first two parts of the book.  More work is definitely still needed, but I've looked at these so much that they're becoming a blur to me, so I think it's time to hand it off.

Here's a partial list of the most important changes:

 - Switch to using `latexdefs.tex` for latex macros and remove all macro definitions from notebooks.
- Changed Advection examples to be interacts where you can change the advection velocity
 - Add a waves+solution plot at the end of the advection chapter
 - Heavy rewriting of the first few sections of Acoustics.ipynb, including some new widgets and a lot of code that probably should be moved out of the notebook.
 - Use seaborn, mostly to set good defaults for matplotlib font sizes.  I haven't applied this to all notebooks; I wanted to get feedback first.
 - Introduce a 'context' variable to indicate whether we are building the PDF, HTML, or just running the notebook.  This can be passed to functions in order to modify what output they produce.
 - Major revision of the intro to part II.  I greatly expanded the discussion of HLL solvers (now includes LF and LLF).
 - Added description of another solver in Burgers_approximate.ipynb
 - Added examples showing the need for an entropy fix in Burgers_approximate.ipynb
- Remove a lot of old material from the README.
- Update the Index to include the current set of chapters.

Some other minor changes that come to mind:
 - Replace some figures in Traffic.ipynb with plots produced by code
 - Some revision of Burgers.ipynb.  Still needs a lot more work.
 - Changed fill_between colors for shallow water plots to be all shades of blue (resolves #127)
 - Add captions to figures without captions.
 - Fix lots of plotting issues
 - Shorten code lines (in some places, not all) so that code doesn't go off the page in the PDF.


ketch:
Okay, I'm really finished now and I believe this is ready to merge.

maojrs:
I just started looking at this. I will do a full read myself and leave you here the comments. I noticed that the new interact plots you added are not quite working for me. They plot image after image in a vertical fashion. I noticed this is resolved by adding plt.show() at the end of the plotting functions that are called by the interact. However, I am not sure if you are having the same issue.

ketch:
> the new interact plots you added are not quite working for me

Can you tell me specifically which ones?  Then I'll fix them.

maojrs:
There are several. For instance, all the interacts in the advection notebook except the last one. The same for the acoustics notebook, except for the interactive phase plane plots.

ketch:
I can't reproduce that, so I'm guessing it's a difference in the libraries we're using.  Can you update to the latest versions of matplotlib, jupyter, and ipywidgets?  Then if the problem persists, tell me what versions you have of each of those.

maojrs:
It seems it is working for Randy, so it is probably my current version configuration. I prefer not to play with it at the moment, though I will update it soon. Anyway, I finished a detailed review up to the end of Part I. I will send this via email since it might be too long for github.

PR #133, opened 2018-05-28
Animation tools fix

rjleveque:
Fixes to `make_html` and `make_rst` to properly load html text.

Added `animation_tools_demo.ipynb` for testing everything.

### DIFF
--- a/animation_tools_demo.ipynb
+++ b/animation_tools_demo.ipynb
@@ -1,0 +1,353 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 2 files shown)


ketch:
This is great; I very much like the manual option for slower interacts and we should incorporate that into some of the widgets in the book.

I guess that a lot of the text in the notebook can also be incorporated as docstrings for the functions.  I sent you a PR with this.

Now if only we could figure out how to avoid rasterizing complicated plots when making jsanimations.

By the way, the notebook says "JSAnimation package from ?".  I'm not sure what is supposed to go in place of the "?".

PR #136, opened 2018-06-06
Add comparison for Euler solvers using PyClaw.

ketch:
This is a cleaned-up version of #135.

### DIFF
--- a/Euler_approximate_solvers.ipynb
+++ b/Euler_approximate_solvers.ipynb
@@ -600,7 +600,439 @@
    "source": [
     "Euler.phase_plane_plot(q_l,q_r,approx_states=states,cons_inputs=True)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Comparison of two approximate solvers with the exact solution\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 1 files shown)


rjleveque:
Thanks @ketch, I like it!  Some comments:

1. There seems to be a block of stuff repeated twice, from the "Comparison of two approximate solvers with the exact solution" (which doesn't seem to show that) followed by a second copy of "Transonic rarefactions and an entropy fix" and "HLLE Solver" sections.

1. There's a `\wave` macro used in (18) that is undefined.

1. You might point out that when rho goes negative in the Roe solution the velocity and pressure blow up with huge values.

1. Should I try to write something on transonic rarefactions?

1. I have mixed feelings about horizontal vs. vertical layout.  Vertical allows one to see how the waves line up between the different plots better and works better on my narrow laptop screen in the sense that one plot doesn't go off the screen to the side, but then with vertical I have to scroll up and down a bit to see all four plots (when I have things magnified to a size where my old eyes can read the text). On balance I think I prefer vertical since it shows more detail and how things line up.

1. By "Second-order modified Lax-Wendroff algorithm" I assume you mean with limiters?  Maybe best to add "... with limiters" to this heading since people might expect L-W to be more oscillatory.

1. In the "High-order WENO + Runge-Kutta" section I get warnings like `euler_1D_py.py:309: RuntimeWarning: invalid value encountered in sqrt` although the plots look fine.

ketch:
I'm closing this since it is replaced by the contents of Euler_compare.ipynb.  I have tried to resolve most of the comments from @rjleveque in #163.  I went with the vertical layout but also made the plots a bit smaller so they should all fit on the screen.

The section on transonic rarefactions is still missing; if you want to write it, that would be great!

PR #137, opened 2018-06-14
WIP: making a pass on Part I

rjleveque:
I'm still working on this, but here are some comments so far...

1. **Introduction.ipynb**

   1. I think the example with oscillatory data is distracting in this
      introduction where we are trying to get across the main idea of the
      Riemann problem.  Maybe just state the limiting results without showing
      the plots?

   1. I added a bit more about other applications early in this notebook.
      Do we still need the list right before the Godunov's method section?

1. **Traffic_flow.ipynb**

   1. I thought the x-t plot in the section "Example: Traffic jam" might be 
      better if `xlim = (-0.6,0.6)` was always set so that it is easier to see that
      the shock speed varies with `ql`.  But then the spacing between
      characteristics to the right keeps changing, which is not optimal either.
      So we should either fix this or take back out the `xlim`.

1. **Shallow_water.ipynb**

   1. Some legends on top of states in phase plane plots.

   2. I changed $q_\ell$ to $q_l$ several places.

   1. Changing the 2-rarefaction u data to -2, 2 as instructed gives
      `ValueError: figure size must be finite not (nan, nan)`
      -1.9, 1.9 is ok.

1. **Shallow_tracer.ipynb**

   1. Include pure Riemann problem with piecewise constant `\phi`.
   
   2. Switched from pyclaw to exact solver.
   1. Still need to add more discussion.

1. **Euler_equations.ipynb**

   1. We write out the 3D Euler equations.  Should we also write out the 2D
      shallow water equations in `Shallow_water.ipynb`?  Or drop these from
      Euler as distracting?

   1. I replaced the deprecated find function in lines like
      `n = find(t > t_traj)`
      with things like:
      `n = np.array([j for j,v in enumerate(t > t_traj) if v])`

      We need to fix other codes in the same way.
      Opinions seem to differ on the best way to do this kind of thing.
      I don't understand why `find` isn't in numpy.


ketch:
>I think the example with oscillatory data is distracting in this
introduction where we are trying to get across the main idea of the
Riemann problem. Maybe just state the limiting results without showing
the plots?

I'm okay with that.  Are you suggesting that we eliminate both plots, or just the second (oscillatory) one?

>I added a bit more about other applications early in this notebook.
Do we still need the list right before the Godunov's method section?

I'm okay leaving the list (it includes things not mentioned elsewhere in the introduction) or removing it.

>I thought the x-t plot in the section "Example: Traffic jam" might be
better if xlim = (-0.6,0.6) was always set so that it is easier to see that
the shock speed varies with ql. But then the spacing between
characteristics to the right keeps changing, which is not optimal either.
So we should either fix this or take back out the xlim.

I like the change you made and I favor the option of keeping it and fixing the issue with the vehicle trajectory density.  The density of those lines will then make more sense in any case.

In Shallow_water.ipynb there seems to be a stray cell with just `widgets.Checkbox`; maybe I left that there by mistake?

> Some legends on top of states in phase plane plots.

Right -- I'm open to proposed solutions, including just explaining the meaning of all lines up front and omitting the legends.

>Changing the 2-rarefaction u data to -2, 2 as instructed gives
ValueError: figure size must be finite not (nan, nan)
-1.9, 1.9 is ok.

That seems to be a bug.  But the proposed change seems fine.

All other changes look fine to me so far.  I have yet to look over the other two chapters.

ketch:
Comments on shallow water w/tracer:

- The description mentions stripes, but the plots don't use stripes (I vote to change the description and keep the plots without stripes)
- The PyClaw stuff was put there a long time ago before we settled on how to do things.  I did it because the images we had from the exact solver at the time looked very bad.
- This chapter feels a little superfluous, since we've been showing \phi since the introduction.  I guess the purpose is to talk about linearly degenerate fields, which isn't included yet.

Comments on Euler:

>We write out the 3D Euler equations. Should we also write out the 2D
shallow water equations in Shallow_water.ipynb? Or drop these from
Euler as distracting?

I guess there's no compelling reason to include them here, since we don't discuss the Riemann problem for the 3D equations.  I'm okay to remove them.  I wonder if it would be worthwhile to include a last chapter of part I discussing the 1D Riemann problem for multidimensional systems like shallow water and Euler?  From a practical point of view this is quite essential.

- Regarding the use of `find`, doesn't `numpy.where` do exactly this?

rjleveque:
@ketch:

> Regarding the use of find, doesn't numpy.where do exactly this?

I don't think so.  `where` returns an array of the same shape as the inputs, while `find` returns a list of the indices where the condition is true, which might be much shorter than the total number of indices in the array.

PR #139, opened 2018-07-01
WIP: making a pass on Part II

rjleveque:
I'm starting to go through Part II.  A few questions so far...

1. **Approximate_solvers.ipynb**

   1. I had to define macros for \wave, \amdq, \apdq for the latex to work
      and I did so in a cell above the first cell that appears.
      Is this already done somewhere else?

      I tagged this cell as "raw", thinking we may need to convert this into
      a "raw nbconvert" cell in order to get it to render properly in the pdf?
      See https://github.com/jupyter/nbconvert/issues/312.

1. **Burgers_approximate.ipynb**

   1. In the transonic rarefaction section, `q_m` is used in several places
      that should possibly be `q_s`?

   1. Should we move some of the code, e.g. the definition of `setup`, to 
      a separate file that is imported?

### DIFF
--- a/Approximate_solvers.ipynb
+++ b/Approximate_solvers.ipynb
@@ -2,6 +2,21 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "metadata": {
+    "tags": [
+     "raw"
+    ]
+   },
+   "source": [
+    "$$\n",
+    "\\newcommand{\\wave}{{\\cal W}}\n",
+    "\\newcommand{\\amdq}{{\\cal A}^-\\Delta Q}\n",
+    "\\newcommand{\\apdq}{{\\cal A}^+\\Delta Q}\n",
+    "$$"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "metadata": {},
    "source": [
     "# Part II.  Approximate Riemann Solvers"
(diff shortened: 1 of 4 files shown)


ketch:
>I had to define macros for \wave, \amdq, \apdq for the latex to work
and I did so in a cell above the first cell that appears.
Is this already done somewhere else?

I created an extension to load a set of macros.  If you update to the latest release of jupyter_contrib_nbextensions you should have access to it; it's called `load_tex_macros`.  The macros themselves appear in the file latexdefs.tex.  If you enable the extension, the macros in that file should automatically be loaded in any notebook you open.  This means that our readers will also need to enable this extension; let me know if you think this is a good approach.

rjleveque:
I seem to have that nbextension activated but it still doesn't work for me unless I add the latex macros to the notebook explicitly.  Not sure what's wrong.

rjleveque:
I just pushed some minor changes to a couple more notebooks.  Some notes on these:


1. **Shallow_water_approximate_solvers.ipynb**

   1. Nice derivation of the Roe averages!  I suppose it is conceivable that 
   the averages could depend on $g$ in some way, but this derivation shows
   there's no need to introduce that complication.

   1. Some places $\lambda_p$ is used and other places $\lambda^p$.
   For consistency with other notebooks I guess we want $\lambda_p$
   but then there is also $\wave^p_r$ for example where a superscript is
   used.

   1. The plots in the Positivity and HLLE sections are missing legends 
   telling exact from approximate.

   1. In the "Entropy and positivity" section of the HLLE Solver, it says
   HLLE does not require an entropy fix and the first example is a transonic
   rarefaction that is presumably supposed to illustrate this.  I'm not sure
   this will be convincing though since it naturally shows that HLLE replaces
   the transonic rarefaction with a single left-going wave, just as Roe does.
   It is only because the intermediate state is much higher than with Roe
   that this ends up being entropy-satisfying when used in a numerical method, 
   I think.   I don't know if there's a good way to explain this?

   1. In general I like this notebook a lot!  Very clear and nice plots.


1. **Euler_approximate_solvers.ipynb**

   1. Spacing in equation (6) is odd, move $f(q)$ to the left?

   1. I added a bit of discussion about the fact that the p-u phase plane
   has one dot for the true middle state and two dots for the approximate
   solution.  Should we start with an example that shows this more clearly?
   In the next example shown this is clearer.

   1. Still need a section on transonic rarefactions.

   1. Still need to decide on horizontal vs. vertical layout of subplots.

PR #140, opened 2018-07-05
Change command-line option in nbconvert template for compatibility with updated pandoc.

ketch:
This fixes the issue that all the text was stripped out of the PDF when using recent versions of pandoc.


ketch:
With this modification, the PDF generates correctly for me.

rjleveque:
Thanks @ketch, works for me too!

PR #142, opened 2018-07-06
Move demo code from traffic notebook into traffic_demos.py.

ketch:
For the moment, `traffic_demos.py` is in the `exact_solvers` folder.  I tried putting it in a parallel folder called `demos`, but then it was more difficult to import.  I think we would need to add an `__init__.py` in the base `riemann_book` folder (so that the whole book is a Python package) and then readers would need to add the folder above `riemann_book` to their python path, or we would need to have Python actually install the package to a system folder.  I wasn't sure that we wanted to add that additional installation requirement.  It seems simpler to have the exact solvers and the demos in separate files in the same folder.  Perhaps we could rename the folder to `solvers_and_demos` or something similar.  In any case, I think it's okay to merge this PR and decide later on the directory structure.

### DIFF
--- a/Traffic_flow.ipynb
+++ b/Traffic_flow.ipynb
@@ -45,6 +45,15 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "If you wish to examine the Python code for this chapter, please see:\n",
+    " - [exact_solvers/traffic_LWR.py](exact_solvers/traffic_LWR.py)\n",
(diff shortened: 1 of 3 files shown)


rjleveque:
I like this!  I added links to the code at the top for those who want to examine it, does this seem like the right place to put it?

ketch:
That seems fine, or we could put it at the end. In any case it's easy to move that cell.

PR #143, opened 2018-07-08
WIP: revert to local JSAnimation

rjleveque:
Because the version built into matplotlib seems to have problems.

- Not all animations play properly in html files
- Plots are greatly pixelated in the animations

I think this reverts back JSAnimation while preserving other more recent updates to `utils/animation_tools.py`.

I ran into problems testing it however -- I foolishly used the same conda environment to try to re-build the clawpack docs to fix the pyclaw issues there, and now the pyclaw Riemann solvers don't import properly any more (apparently due to numpy having been updated when I installed nbsphinx).  Such fun!

I'll work more on this later...

### DIFF
--- a/Nonconvex_scalar.ipynb
+++ b/Nonconvex_scalar.ipynb
@@ -4,9 +4,47 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


rjleveque:
This branch and #144 got mixed up.  I'm closing both and opening new ones.

PR #144, opened 2018-07-09
Nonconvex_scalar updates

rjleveque:
I removed most of the code from `Nonconvex_scalar.ipynb` and added much more discussion and some additional example plots.

### DIFF
--- a/Nonconvex_scalar.ipynb
+++ b/Nonconvex_scalar.ipynb
@@ -4,9 +4,47 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Nonconvex scalar Conservation Laws and the Osher Solution\n",
+    "# Nonconvex scalar Conservation Laws"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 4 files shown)


rjleveque:
I just added commits to change how `riemann_tools` is imported,  after a couple failed attempts, but now I can't figure out why it did seem to be working according to travis before I "fixed" it.  But it didn't work on my laptop when I started a new session.

The original code I replace was this, which also appears e.g. in `exact_solvers/traffic_LWR.py`:

```
sys.path.append('../utils')
from utils import riemann_tools
```

I don't see why this works since `utils` is not a subdirectory of `utils`, so it seems like it's `..` that should be on the path, not `../utils` for this to work.  It maybe worked because `..` had already been added to the path by something else, but I'm not sure where.

I tried to clean this up with a check so that `sys.path` doesn't get the top directory appended more than once.  Maybe this is not so important and the code would be clearer without this check.

rjleveque:
Closing this and #143 and opening new ones.

PR #146, opened 2018-07-11
Revert JSAnimate

rjleveque:
Second attempt and untangling mixed branches.

Also change some figsizes for better resolution animations.

It also helps to set
```
mpl.rcParams['figure.dpi'] = 100
```
in each notebook using this, I think.


rjleveque:
I've been testing and seems ok.

PR #148, opened 2018-07-21
Burgers NB

maojrs:
As promised, here is an improved version of the Burgers' NB.
Summary of changes:
- Incorporated Randy's comments/changes
- Removed code from NB into files.
- Separated (and improved) code into burgers.py and burgers_demos.py 
- Added links at beginning of notebook to code files
- Added subsection about entropy condition (thought it was needed since it is just briefly mentioned in the traffic NB)
- Incorporated a bump and a three state initial condition examples. I think these make the NB much more interesting. However, I used pyclaw to produce the animations. I can't remember if we agreed to have pyclaw as dependency or not.
- General review of the text and wording.
- Added `plt.tight_layout()` in `plot_riemann` in `utils/riemann_tools.py`. This improved the appearance of the burgers NB plots (horizontal layout), though I hope it doesn't ruin other plots. Other NBs seemed fine on my computer.


rjleveque:
@maojrs: I like it!  I made a few minor suggestions and added a short section in my PR to your branch.

For the last example, should we also show the characteristics for the merging shocks case?  I guess we could for for the shock overtaking the rarefaction too if we did a contour plot in the x-t plane of the pyclaw solutio at many times, but maybe not worth spending time on that now.

maojrs:
@rjleveque Thanks for the comments and changes! I think having the characteristics in the last example would be nice too. I will take a look into it, but first I will finish the acoustics notebook upgrade.

rjleveque:
The new bump example gives me an error:

```
AttributeError: module 'clawpack.riemann.burgers_1D_py' has no attribute 'rp1'
```
I'm using commit clawpack/riemann@45004b17362 which has been working for other notebooks. Is there an update?

PR #149, opened 2018-08-01
Acoustics

maojrs:
Finally the acoustics one as well! Here is a summary of changes:
- Implemented **NEW full interactive solution w/phase plane** at end of notebook.
- Review and rewrote parts of the text/code to improve clarity.
- Moved code from NB into `acoustics.py` and `acoustics_demos.py`.
- Moved code from `interactive_pplane.py` into `acoustics_demos.py`.
- Removed code from `acoustics.py` and `interactive_pplane.py` and moved it to `acoustics_heterogeneous.py` and `acoustics_heterogeneous_demos.py` (also modified notebook `acoustics_heterogeneous.ipynb` to maintain previous functionality).
- Added links at beginning of notebook to code files.


maojrs:
Thanks for checking it out! I modified the time domain in `char_solution_interactive`. 

Regarding the figure size, I made it smaller in all of the figures. I did try to use a global variable like in `Traffic_flow.ipynb`, but then I got skinny looking plots and overall inconsistent looking figures since some figures require a slightly different figure size. I guess this is a global issue in all the notebooks, so maybe we should discuss possible solutions. One possibility would be to have a range of figure sizes that we know works well or maybe to use a scale factor that can scale all figures proportionally.

PR #153, opened 2018-08-05
Move code out of notebooks

ketch:


ketch:
I think this is ready to merge now.  Sorry that it includes so many miscellaneous things; I just wanted to fix small things when I noticed them so that I wouldn't forget.

PR #156, opened 2018-10-01
fix figsizes in  Nonconvex_scalar by loading matplotlib in separate cell

rjleveque:
Also I added a comment that the last example doesn't work in the html version:  I think because of multiple sliders not converting to jsAnimation, this only works for a single time-slider.  Maybe we should figure out a way to handle other cases?

Hmm, I take it back.  For some reason the last time I created the new html file it did work, though it didn't previously. See http://www.clawpack.org/riemann_book/html/Nonconvex_scalar.html

### DIFF
--- a/Nonconvex_scalar.ipynb
+++ b/Nonconvex_scalar.ipynb
@@ -61,6 +61,15 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


ketch:
Looks like the last example works in the linked version, but only has the slider for t (I guess that's what you said).  I'll go ahead an merge this.

PR #160, opened 2018-10-03
Minor edits to Burgers' chapter.

ketch:
@maojrs I went ahead and merged your Burgers chapter update, and then read through it.  In this PR I've made some small changes.  

I think some additional bigger changes should be made.  The main idea behind these changes is to take advantage of our unique medium.  The concepts involved in this chapter have been explained many times elsewhere (including in Randy's books) and we are not going to do a better job with just words to explain these concepts, so I would prefer not to spend a lot of words on them.  However, we have the advantage of interactive illustrations and we should use those wherever they can add something.

- The first figure should be an interact, with a slider for the time.  It's just begging for it.  Also, I think it's helpful to remove the axis tick labels in this figure.
- The same goes for the second figure.  There could be a slider for the location of the shock and an indication of how much mass is on each side.
- The third figure would also be nicer as an interact.
- I found the section on weak solutions confusing, since you introduce a discontinuous function $\phi$ and then immediately differentiate it.
- It would be really nice to plot more characteristics in the later examples (where PyClaw is used).


ketch:
> Do you have any better idea of how to plot all the charactersitics with pyclaw?

One thing that would be very easy but a little hacky: instead of making the solution constant on intervals, make it vary by just a little (e.g. 10^-5).  This won't change the solution noticeably, but you can then place contour lines precisely in the right intervals to get more characteristics.

Another option would be to numerically integrate the ODE t'(x)=1/q(x,t) for several initial values of x.

ketch:
By the way, I forgot to say that I think overall this chapter is becoming quite nice.  And thanks for going over my other PRs!

maojrs:
@ketch perfect! I just wanted to double check, then I will give it a try to make those interactive and see how it goes. Also thanks for the suggestions for the pyclaw characteristics.  Lets leave this pull request open, and I will push the changes directly here.

ketch:
@maojrs I think it looks great now!  

I changed "unphsyical" to "multivalued" since for water waves, overturning is actually physically correct up to a point.  The equal-area rule interact could be made even nicer by shading the areas and showing their size (numerically) but I don't know if it's worth the effort.  I'll go ahead and merge this.

maojrs:
Perfect!

PR #165, opened 2018-10-20
Improvements acoustics

maojrs:
This pull requests makes some improvements to the acoustics notebook and code:
- Minor text improvements.
- Eliminates plot repetition by merging the Riemann plotter with the phase plot.
- Merged interactive phase plane plot into simple example section.
- Added pyClaw based example for bump initial condition w/reflective walls.


maojrs:
Thanks! I added the js widget I made instead of the python based widget I had. I used the IFrame comman, and it seems to be working well on my computer (with the width and height parameters I used). Let me know if it looks well on yours too. 

I also had to push a new version of 'html/phase_plane_acoutics_small' without the title. I guess we should decide if we even need the title in those or if we embed them directly into the notebooks.

maojrs:
Thanks! I added the js widget I made instead of the python based widget I had. I used the IFrame comman, and it seems to be working well on my computer (with the width and height parameters I used). Let me know if it looks well on yours too. 

I also had to push a new version of -html/phase_plane_acoustics_small- without the title. I guess we should decide if we even need the title in those or if we embed them directly into the notebooks.

maojrs:
Thanks! I added the js widget I made instead of the python based widget I had. I used the IFrame comman, and it seems to be working well on my computer (with the width and height parameters I used). Let me know if it looks well on yours too. 

I also had to push a new version of 'html/phase_plane_acoustics_small' without the title. I guess we should decide if we even need the title in those or if we embed them directly into the notebooks.

maojrs:
Thanks! I added the js widget I made instead of the python based widget I had. I used the IFrame comman, and it seems to be working well on my computer (with the width and height parameters I used). Let me know if it looks well on yours too. 

I also had to push a new version of 'html/phase_plane_acoustics_small' without the title. I guess we should decide if we even need the title in those or if we embed them directly into the notebooks.

rjleveque:
It's great having the widget inline for the notebook, but it's not going to render properly in the html version as interactive html.  It does show up as a static figure, shown below (once I copied the file over to the `build_html` directory), but we might also want to include a link to the interactive version for people reading the html.  I think that's why in the Introduction we decided to just include a link rather than embedding the html.
pplane

maojrs:
I wrote an interactive version based on jupyter widgets (much slower than the JS one), so we can use that one instead and leave a link to the html one. You can see it in the original commit of this pull request. Either way is ok with me. Let me know which one you think is best, and I'll push the changes here.

PR #168, opened 2018-10-27
Some wording and spacing changes in Burgers.ipynb

rjleveque:
I really like the interacts in this notebook!

I wonder in this and other notebooks whether we plan to put the blocks of code setting things up at the top, or after the introductory material for the chapter?  I thought we had decided on the latter, e.g. as in `Nonconvex_scalar.ipynb`, but in the `Burgers.ipynb` and some others it cIomes first.

I replaced `Eq. \ref{eq:Burgersintclaw}` with `(\ref{eq:Burgersintclaw})` to be consistent with what is done some other places, although I see we also use `\eqref{eq:...}` some places, which seems to give a spurious period before the equation number (e.g. in `Acoustics.ipynb` both the above forms are used in different places).

I made a few other suggested changes to wording in this PR and removed some white space from `interact(...` commands that were too wide for my window.

I'm happy with this notebook!

### DIFF
--- a/Burgers.ipynb
+++ b/Burgers.ipynb
@@ -147,7 +147,7 @@
    "outputs": [],
    "source": [
     "interact(burgers_demos.shock(), t=widgets.FloatSlider(value=0.0,min=0,max=1.0),\n",
-    "                    which_char=widgets.Checkbox(value=True,description='Show characteristics'));"
+    "         which_char=widgets.Checkbox(value=True,description='Show characteristics'));"
    ]
   },
   {
@@ -201,7 +201,7 @@
    "outputs": [],
    "source": [
     "interact(burgers_demos.rarefaction(), t=widgets.FloatSlider(value=0.0,min=0,max=1.0),\n",
-    "                    which_char=widgets.Checkbox(value=True,description='Show characteristics'));"
+    "         which_char=widgets.Checkbox(value=True,description='Show characteristics'));"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


ketch:
Thanks; these look good.

>I wonder in this and other notebooks whether we plan to put the blocks of code setting things up at the top, or after the introductory material for the chapter? I thought we had decided on the latter, e.g. as in Nonconvex_scalar.ipynb, but in the Burgers.ipynb and some others it cIomes first.

I would rather have some boilerplate at the beginning of every notebook that I know I can skip, rather than an interruption in the middle.  But we had a long discussion about this before and you both preferred to put it lower down, so let's stick with that (unless you've changed your mind).

> I replaced Eq. \ref{eq:Burgersintclaw} with (\ref{eq:Burgersintclaw}) to be consistent with what is done some other places, although I see we also use \eqref{eq:...} some places, which seems to give a spurious period before the equation number (e.g. in Acoustics.ipynb both the above forms are used in different places).

We definitely want parentheses around equation numbers.  I don't see the spurious period (looking at places where \eqref is used in Acoustics.ipynb).  @maojrs do you?

rjleveque:
Another possibility might be to put a lot of the boilerplate in a separate file?  I took a stab at how this might look in #172.

Regarding the period around references, here's how it looks for me...
eqref

maojrs:
I don't have the problem with the additional dots when using eqref, though I guess we can stick to /ref to avoid any problems. If so, I will take a pass to the acoustics notebooks and make sure all the equation references are /ref instead of /eqref and that they have parentheses.

Regarding the setup code, if we are thinking of the printed version, maybe it is better after the intro, so it is not the first thing it is shown. However this cells are hidden in the printed version right? In that case, I agree with David that it is better to have it before. I'll comment more on #172.

maojrs:
Were the changes here already introduce in pullrequest #194 ?

ketch:
>Were the changes here already introduce in pullrequest #194 ?

I believe the "files changed" page always shows a comparison with the current master, so there are changes here that still need to be merged.  Let me see if I can resolve the conflicts and merge this.

ketch:
Never mind; after looking at the conflicts, I guess I was wrong.  Everything here has been superseded by changes we already merged.

PR #169, opened 2018-10-27
Some suggestions to Euler_compare.ipynb

rjleveque:
I think this is a useful notebook.

For the Sod problem it is stated that the "fine grid" solution can be viewed as the exact solution, but as computed (especially in the first example where the clawpack solver is used with only 1000 points) the contact discontinuity is quite smeared out, which could lead to confusion on the part of the reader.  I tried changing to sharpclaw with 2000 points and it looks a bit better, maybe go to an even finer grid?  Or hardwire the exact solution since we have it?

I added some comment about the fact that a single Riemann problem is solved numerically to perhaps avoid confusion with the exact Riemann problem.

The initial data used for the blast wave problem seems to be specified as
```
q_l = Euler.conservative_to_primitive(1.,0.,1.)
q_r = Euler.conservative_to_primitive(1./8,0.,1./10)
```
but then these values are ignored in the `blastwave` function invoked in
```
roe_bw = blastwave(q_l,q_r,riemann_solver='Roe')
```
since this problem has 3 initial states that are hardwired into the function definition.  Am I missing something?  Can we remove `q_l, q_r` from this example?

It would be good if the plots could be titled something like `Density at t = 0.038` so the reader knows which variable is being plotted, and for comparison with e.g. Figure 15.5 in FVMHP.  Maybe add a bit more discussion of the blast wave problem?

Would it be useful to also plot u and p for these examples?

Many of the examples in this notebook threw a warning such as,
```
/Users/rjl/clawpack_src/clawpack-rbook2/clawpack/riemann/euler_1D_py.py:298: RuntimeWarning: invalid value encountered in sqrt
  rhsqrtl = np.sqrt(q_l[0,...])
```

### DIFF
--- a/Euler_compare.ipynb
+++ b/Euler_compare.ipynb
@@ -11,10 +11,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "How do these solvers impact the solution accuracy when used within a finite volume discretization?  To investigate, we will use them within [PyClaw](http://www.clawpack.org/pyclaw/) to solve two standard test problems for one-dimensional compressible flow:\n",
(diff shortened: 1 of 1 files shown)


ketch:
Thanks!  I will merge this and work on cleaning up the other issues you've pointed out.

PR #170, opened 2018-10-27
Some changes and some things to fix in Shallow_tracer.ipynb

rjleveque:
I added a bit to the discussion of the contact discontinuity.  Do we want to include the general definition of a linearly degenerate family and work out `\nabla \lambda \cdot r`?

I also added a few notes about some things that don't work quite right, including a `\ref` that I can't figure out what's going wrong.


ketch:
Thanks.  I agree that all the things you've pointed out in the notebook need to be fixed.  I always had the impression that this chapter was your baby; I only did some work on it because it was still needed and I knew you weren't going to get to it right away.  So if you'd like to fix those things then that would be great.  You can also choose whether to expand to a general discussion of linearly degenerate fields here (currently that appears in the next chapter).

If you can't get to those things then I will do it at some point.

rjleveque:
Yes, I'll work on this one some more.  I forgot that we describe linear degeneracy more in the Euler notebook, thanks for the reminder.

If either of you figure out what's wrong with that one `\ref` that gives (???), let me know.

ketch:
The equation numbering issue seems to be related to use of `split`.  I removed that and the reference works.  That doesn't really solve the problem, since what we'd like to use here is split.  I tried subequations and it doesn't work at all.

rjleveque:
I modified `exact_solvers/shallow_water.py` to plot the tracer wave too, then I noticed it seems to always give a checkbox for plotting the tracer characteristics in the phase plane even when `plot_riemann_SW` is called with `tracer = False` as in the `Shallow_water` notebook.  I guess that's good since it's nice to see the particle paths also in that case, but the checkbox comes out with a different label when called with `tracer = True` and I don't see where this is happening.

I also updated the `Shallow_tracer` notebook with some recent changes to master.

rjleveque:
Now particle paths are labeled as such in `Shallow_water` for x-t plots where the solution is physically correct, and suppressed for examples with non-physical solutions.

I also fixed up the `Shallow_tracer` notebook, and added a bit more discussion.  I think this could be merged.

PR #171, opened 2018-10-28
Improve Euler_compare.ipynb based on suggestions from @rjleveque.

ketch:
This addresses some of the issues raised in comments on #169.

Use exact solution for Sod shocktube.
Add titles to plots.
Remove unused ql, qr arguments to blastwave function.


rjleveque:
In the blast wave plots I'd suggest changing 
```
    plt.title('Density at t={:.2f}'.format(t))
```
to format ` t={:.3f}` so more digits print.

There is a disagreement between these results and Figures 15.5, 15.6 in FVMHP which show t=0.038.  The notebook results look similar to the figures at t=0.0304 (frame 8) rather than at t=0.038 (frame 10).

ketch:
Good idea about the digits.

>There is a disagreement between these results and Figures 15.5, 15.6 in FVMHP which show t=0.038. The notebook results look similar to the figures at t=0.0304 (frame 8) rather than at t=0.038 (frame 10).

You're right, and this is really puzzling.  Do you have a copy of the source for that example?  I can no longer find it on the internet.

ketch:
Of course, as soon as I posted that I remembered where to find it:

http://depts.washington.edu/clawpack/clawpack-4.3/book/chap15/wcblast/www/index.html

Anyway, I will track this down and update.

ketch:
Somehow the BCs had been set to extrap instead of wall.  I fixed that and it looks right now.  I made a few other improvements, including adding a description of the blast wave problem.

rjleveque:
This looks good, thanks for fixing the WC blast problem so it now looks right.  Perhaps it's worth pointing out that this notebook shows only the 1st order Godunov solution, corresponding to the top plots in Figure 1.5, and that much better results can be obtained with high-resolution methods?

I'm going to go ahead and merge this for now...

ketch:
>Perhaps it's worth pointing out that this notebook shows only the 1st order Godunov solution, corresponding to the top plots in Figure 1.5, and that much better results can be obtained with high-resolution methods?

What do you mean?  The code running here is the classic Clawpack algorithm, including high-resolution correction terms.  What Figure 1.5 are you referring to?

rjleveque:
Sorry, I meant Figure 15.5 in FVMHP.  I thought the plot in the notebook looked much more smeared out, but it was at least partly due to the different aspect ratio.  But not entirely.  I just ran the example in `classic/examples/euler_1d_wcblast` with N=400 and 4000 (red and black respectively in the plot below).  Aren't these the values used in the notebook?  The classic code used the Roe solver, and the MC limiter with CFL=0.9.  The plot looks significantly different than in the notebook.

wcblast_classic
wcblast_notebook

ketch:
The difference was the limiter; I think PyClaw defaults to minmod (perhaps we should change that?)

Here is the figure after switching to the MC limiter:
screen shot 2018-11-13 at 9 06 49 am

Thanks for noticing this.  It's a good idea to use MC here for consistency with your previous book, and it also increases the difference between the solutions given by the two different Riemann solvers.  I'll open a new pull request.

PR #172, opened 2018-10-28
try running a config file at start of notebook

rjleveque:
Experiment to see how it looks if we move some of the boilerplate to a separate `config.py` file. Maybe we could have a common file for all notebooks.

What do you think?

### DIFF
--- a/Nonconvex_scalar.ipynb
+++ b/Nonconvex_scalar.ipynb
@@ -8,6 +8,17 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "%config InlineBackend.figure_format = 'svg'\n",
+    "%run Nonconvex_scalar_config.py"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


rjleveque:
I modified this to use `exec` instead of `run` since `run` doesn't work when converting to pdf or html.  And also moved the `%matplotlib inline` back to a separate cell so the figure sizes come out properly.

ketch:
I'm torn -- it would be nice to have the notebooks look a bit cleaner, but using exec is not very appealing..  The cell is already hidden in the PDF, so it's only there in the notebook.  And it justmeans is they have to hit shift+enter once and scroll their eyes down a couple of inches.

It may also be good if the reader knows what packages are being imported.  In case they are missing a dependency, I think it will be easier to figure that out and install the missing package if we keep the code in the notebook.  So I'm not sure I would favor this change.

ketch:
>Good points. Regarding missing packages, one option would be to have in introductory configuration notebook we suggest people run first that imports all the requirements and contains some tips to help people out if something doesn't import. Then once they have seen what all is required and confirmed they have it, in the later notebooks we could use exec on a config.py?

That should be fine, but I suspect some people will just find a notebook (via Google) and download it from Github without working through the book in order.  Of course we don't need to cater to them.  I'm just not really convinced that the existence of the import cell is a problem that needs to be solved.  Especially if we remove old imports that we don't need anymore.  For instance, in this notebook we can delete the imports of seaborn, matplotlib, and riemann_tools.  Since we still need a cell with matplotlib inline, then there are just 3 more lines of code we're dealing with here.

ketch:
I'm going to do a pass over all the chapters and take out obsolete imports.  Let's see what's left.

maojrs:
I think the notebooks of course look nicer without all the import cells. It would be also nice if the notebooks are standalone an can be downloaded separately. However, in our current setup that wouldn't work because all the demos and exact solvers need to be downloaded as well, so they need to install the whole book regardless

On the other hand, I think for the interactive mode it is better to keep them in the notebook. I think jupyter notebooks are becoming quite popular and people are familiar with having the import statements in the beginning, including myself.

Bottom line, I am not sure what is better, but I am leaning towards leaving the import statements in the notebooks.

rjleveque:
I agree that #174 removes the need for this approach.

PR #173, opened 2018-10-28
Cannot start a line with + sign in markdown

rjleveque:
Equation (4) wasn't  rendering properly in the html version and I finally figure out it was because one line of the markdown cell started with a + sign, and the line was interpreted as a bullet point even though it was in the middle of a latex align environment.


ketch:
That's unfortunate.  Thanks for tracking this down.

PR #174, opened 2018-10-29
Remove unused imports in all notebooks.

ketch:
Motivated by https://github.com/clawpack/riemann_book/pull/172, I tried to cut down the number of imports in our notebooks.  Since we have moved so much of the code out of the notebooks, we only need a few imports now.

I removed all the code that set things like figsize and font sizes for matplotlib inside the notebooks.  To my eye almost everything still looked correct.  Where there were problems, I added code to the plotting functions to set the figure size.  I think that we should put this kind of code down in the plotting functions rather than in the notebooks; what do you guys think?

I also removed the use of seaborn since newer versions of matplotlib use a styling that is almost identical to the seaborn style we were using.  This means we no longer have seaborn as a dependency.

Finally, I realized that we have shoved some of our calls to `interact` down into code that is not in the notebook.  I this means that our approach to using different interact functions when making the PDF or HTML versions will break there.  We should check.


rjleveque:
I like this!  Happy to do this rather than #172.

PR #175, opened 2018-11-01
Move more code to files

ketch:
This PR moves some code into files and consolidates a lot of duplicated code into functions.  There's still much more code in the approximate solvers notebooks than in the notebooks of Part I, but I think this is okay.


ketch:
I'm going to go ahead and merge this in order to keep working, since there have been no comments in 4 days.

PR #176, opened 2018-11-05
Remove latex macro and add transonic rarefaction example in euler approx notebook.

ketch:
I decided not to write a lengthy section on entropy fixes for Euler since it is so similar to SW where I did write a longer section.  Instead, I've just added an example of a solution that would need an entropy fix along with a brief discussion.

### DIFF
--- a/Euler_approximate.ipynb
+++ b/Euler_approximate.ipynb
@@ -111,7 +111,7 @@
     "\n",
     "Solving the system of equations\n",
     "\\begin{align}\n",
-    "q_r - q_l & = \\sum_{p=1}^3 \\wave_p = \\sum_{p=1}^3 \\alpha_p r_p\n",
+    "q_r - q_l & = \\sum_{p=1}^3 {\\mathcal W}_p = \\sum_{p=1}^3 \\alpha_p r_p\n",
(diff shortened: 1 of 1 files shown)


rjleveque:
I notice that the example given for a transonic rarefaction wave also illustrates that the Roe solver loses positivity, although this is discussed separately later in the notebook.  

Maybe it would be better to use this data, which is the same as the second example in the notebook, but with the velocities shifted by -1 to give a transonic rarefaction?

```
left_state  = State(Density = 0.1,
                    Velocity = -1.,
                    Pressure = 0.1)
right_state = State(Density = 1.,
                    Velocity = 0.,
                    Pressure = 1.)

states = compare_solutions(left_state,right_state,solvers=['Exact','Roe'])
```

ketch:
Good point.  I hadn't even noticed the negative density there.  I liked that in that example the Roe solver sent nothing to the right (even though part of the rarefaction should go right), so I found a positive example that shares that property:

```
left_state  = State(Density = 0.1,
                    Velocity = -2.,
                    Pressure = 0.1)
right_state = State(Density = 1.,
                    Velocity = -1.,
                    Pressure = 1.)
```

PR #177, opened 2018-11-11
Fix eqref in markdown cells.

rjleveque:
changed \eqref{label} to (\ref{label}) in all the current chapters
using the script fix_eqref.py


ketch:
Thanks for also automating this so we can easily fix any future mistakes.

PR #178, opened 2018-11-11
WIP: Fix pdf problems

rjleveque:
I'm still working on fixing various things that were broken in the pdf version.

The `make_pdf.py` script now changes all the code in `exact_solvers` to replace `from ipywidgets import interact`  by  `from utils.snapshot_widgets import interact`.   

Note that both `exact_solvers` and `utils` get copied into `build_pdf` and it's only the code in `build_pdf/exact_solvers` that gets modified.

Maybe there's a better way to set the path in all these so that import statements `from utils...` work properly?  I always have problems with relative imports.  

Most of these are working except the one in `acoustics_demos.riemann_plot_pplane` so for now I put it in a try-except block.  

The latex file mostly compiles although it complains about a few things I'm still tracking down.

Lots of code needs to be broken into more lines in the notebooks so that it fits on the page in the pdf.


maojrs:
I'm trying to fix the acoustics_demos.riemann_plot_pplane snapshots, but now I can't seem to get make_pdf.py work properly. I am getting the error: No such file or directory: 'riemann.tex', so it seems bookbook is not generating the tex file. Am I missing something?

I also noticed that it requires the python2 kernel, is this neccesary? Or can it all simply run with python3?

ketch:
It should run with Python 3; I hard-coded that "python2" at one point because I was having an issue with different parts of the code calling different pythons on my machine.  I think I committed it by mistake.  The later part (where python3 is explicitly used to run bookbook) will only run with python 3.

The file `riemann.tex` is created by bookbook, so if you don't have it then bookbook is probably not executing properly.  I would look for an earlier error in the output.

rjleveque:
For me it was originally failing because some notebooks had an empty cell (or latex macros cell) at the top, and bookbook requires that the first cell start with a single `# Title` to set the chapter title, so bookbook failed before getting to the stage of making riemann.tex.  But I think the version in my PR had these fixed.

maojrs:
I found the error. For some weird reason if I simply run make_pdf.py, bookbook generates an error and doesn't create riemann.tex. However, if I simply run bookbook directly from the command line, it works. I have no idea why this could be happening, but if it works for you, I'll focus my time on getting the other snapshots to work.

ketch:
@maojrs Clawpack (i.e. PyClaw) runs fine in Python 3.6 -- we actually test it on that version of Python.  If you have problems running PyClaw with Python 3.6 you should raise an issue.

maojrs:
@ketch thanks, I'll check it. I might just need to update clawpack in my conda environment.

maojrs:
Fixed and sent a pull request to this branch!

ketch:
I'm working on this now and noticing that the call to bookbook  takes a very long time (2 minutes or more).  I remember it being much faster previously.  Are you guys seeing the same?  Do you have any idea what changed?

rjleveque:
If this looks ok, I suggest merging it in and building on it.  Note this incorporates the new `SIAMbook2016.cls` file received 29 March 2019 with wider text. 

A few things that are broken, to be fixed in a new PR perhaps...

 - `Burgers.ipynb` has an interact that's not working, `burgers.riemann_solution_interact()`
 - Part II notebooks have many code lines that are too long and should be broken up.

exact_solvers/acoustics_demos.py
```diff
@@ -73,9 +79,12 @@ def char_solution_interactive():
 
     mainwidget = interact(char_solution, t=twidget, rho=rhowidget, K=Kwidget);
 
-    mainwidget.widget.close()
-    display(interact_gui)
-    display(mainwidget.widget.out)
+    try:
```
ketch: I guess the reason for putting this in a try/except block is that when it is run in a certain mode then this code raises an exception and doesn't need to be run.  Doing this has gotten me into trouble in the past; it would be nice to catch any exceptions other than the one we expect by writing `except <particular exception type>:`.

exact_solvers/traffic_ramps.py
```diff
@@ -1,8 +1,10 @@
+import sys, os
+top_dir = os.path.abspath('..')
+if top_dir not in sys.path:
+    sys.path.append(top_dir)
```
ketch: Why is this necessary now?  Since it adds boilerplate code to so many files and messes with the user's PYTHONPATH, it would be nice if we could avoid it...

test.py
```diff
@@ -34,8 +34,12 @@ def _notebook_run(path):
 def run_tests():
     doctest.testmod(utils.riemann_tools)
 
+    # notebooks to ignore in tests:
+    ignore_notebooks = ['Make_html_animations.ipynb']
```
ketch: Why do we not test this notebook?

ketch:
Thanks!  I can go through Part II this week.

I added a few comments; if you prefer I can make them issues and then merge this immediately.

exact_solvers/traffic_ramps.py
rjleveque: I guess it works ok without this now.  For some reason I was having a lot of relative import problems earlier.  I'll try removing all this and test it all out again.  Stay tuned...

test.py
rjleveque: It's just run to make animations to post on the webpage for the html version, not needed by users.  It's fine to test too I guess.

exact_solvers/acoustics_demos.py
rjleveque: I don't remember now if I added this or why.  It's fine to improve the code to catch the exception if you want.

maojrs:
Thanks for going over all of these. I can check the `burgers.riemann_solution_interact()` and try to fix it.

exact_solvers/traffic_ramps.py
rjleveque: Yes it works ok and I've removed all these lines in my latest commit.

rjleveque:
I'm not sure why `Burgers` is not passing travis in Python 3.  Before restarting the build it died earlier in Python 3 so I'll try again later, maybe a travis issue.

ketch:
I restarted the build again and it passed.  Must have been a Travis issue indeed.  I'll merge this now.

PR #184, opened 2019-04-04
Shorten long lines of code in part 2

ketch:


ketch:
I think this is ready to merge.  It makes all the code in Part 2 fit within the margins of the PDF.

ketch:
If there are no objections, I will go ahead and merge this tomorrow.

rjleveque:
Sorry, got busy with other things, but I'll take a quick look today...

rjleveque:
Looks good!  A few comments/suggestions:

**Burgers_approximate:**

1. Should we provide links to `exact_solvers/burgers.py` as we do in the Burgers notebook?  Here we include much more code directly so maybe not needed.  Also, should we somehow link to `clawpack/riemann/.burgers_1D_py`?   (Same question applies to other notebooks in Part II.)

2. Regarding your query, I don't think I've ever tried the HLL solver for Burgers', but probably someone has.  Remove this query?

**Shallow_water_approximate:**

1. I like your simplified derivation of the Roe average states, here and in Euler, but should we also give a reference to the more general (?) parameter vector approach from Roe's original paper?

2. For the Roe solver with the entropy fix, also show PyClaw solutions to illustrate that this solution gives good results in the context of a numerical method? (Even though the Riemann solution itself still looks very different from the true solution with discontinuities rather than rarefactions.)  I guess this is covered in the last notebook of Part II (for Euler, anyway), so maybe just refer forward to that notebook rather than showing an example?

3. For  Transonic rarefaction example a, the true solution shows an interesting bump in hu in the rarefaction wave between q_l and q_m.  It might be clearer to the reader why it looks this way if the next cell In[20] used `yaxis = 'hu'` rather than `u`? But all the other phase plane plots use `u` so maybe this would be confusing.

**Euler_approximate:**

1. Equation (12) for $\hat H$  is too long in the pdf.

**Euler_compare:**

1. The figures are too wide on my browser, and wider than in other notebooks.

1. In the discussion of each example, perhaps suggest that the reader adjust the value of `N` in each solver and re-execute the cell.  Perhaps define `N = 50` at the top of the cell and use this `N` in each solver so it's easy to change?  Or make this a widget?

rjleveque:
I just noticed another problem: in `Euler_compare` the pdf version only shows the solution at `t=0`, rather than two times.

ketch:
I implemented your suggestions in the Burgers chapter: linked to `exact_solvers/burgers.py` and `clawpack/riemann/burgers_1D_py.py` and removed my query.  I also shortened the PyClaw script by removing some unneeded options.

ketch:
In the shallow water approximate notebook, I added a link to Roe's paper and chapter 15 of FVMHP that discuss the parameter vector approach.  I included a second h-hu state plane plot for the example with the bump, and a comment about this.

I agree that it would be nice to include some pyclaw simulations in the shallow water notebook.  I will do this in a separate PR.

ketch:
In the Euler approx notebook, I added links to the exact solver and the Clawpack solver, and broke up the equation that was still too long.

The euler_compare notebook still needs some work, which I will do in a subsequent PR.

PR #186, opened 2019-04-17
fix figure sizes in Nonconvex_scalar.ipynb 

rjleveque:
And clean up text and cross references a bit.  

In particular I changed references such as `[Burgers.ipynb](Burgers.ipynb)` to `[Burgers](Burgers.ipynb)` to be consistent with what we do elsewhere and I think it looks a bit nicer in the notebooks.

There was also a hidden cell that should not have been.

### DIFF
--- a/Nonconvex_scalar.ipynb
+++ b/Nonconvex_scalar.ipynb
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


ketch:
Have you checked that those links still get converted correctly in the pdf?  I think they may not.

rjleveque:
Yes, it seems to work fine in both the pdf and html versions.  But looking at other notebooks I see we have used both styles and both seem to work, so we should decide which we prefer.

The changes I made to `Burgers` and `Traffic_flow` in #187 also include dropping the `.ipynb`.

ketch:
What I observe is this:

- If .ipynb is included in the link text, then in the PDF it appears as "Chapter X", where X is determined automatically by Latex, with a working hyperlink to that chapter in the PDF.
- If .ipynb is removed, it appears in the PDF with whatever the link text was and is a link to the .ipynb file, rather than to a location in the PDF.

rjleveque:
Closing this since we should not make the changes to other notebooks in this PR.

I will fix the figsize issues in a new PR.

Issue #185, opened 2019-04-04
Burgers interact


maojrs:
I took a look at the interact not working in burgers. The problems seems to be this line:

`from utils.snapshot_widgets import interact`
should be
`from ipywidgets import interact`

If we use the first one, we get the snapshots as we want right? If we use the second one then we get the interacts. If I remember correctly the script to produce the pdf changed all the `from ipywidgets import  interact` to `from utils.snapshot_widgets import interact`. Maybe we should add to the script to revert everything once the pdf is generated.

I am happy to do that but just want to check with you guys since I didn't write that part, and I don't want to break anything else. 

I also noticed an additional parameter was added to `burgers_demos.multivalued_solution` and `burgers_demos.shock_location` To fix the interact. I don't think this was necessary, we just need to change the import statement on the top of the notebook. 

Should I just revert these changes to the original or am I missing something?

maojrs:
Maybe the confusion comes from having interact being imported inside the notebook and then also in `exact_solvers/burgers.py`?

rjleveque:
I think `burgers.py` was the only file in `exact_solvers` that imported from `utils.snapshot_widgets`, must have been a mistake.  I replaced that by the usual import and now the notebook works fine.  Thanks!

PR #187, opened 2019-04-18
Burgers/traffic swap

rjleveque:
I talked to @ketch yesterday about possibly swapping the order of these two notebooks to put `Burgers` before `Traffic_flow` since I find it more natural to teach them in this order.  This PR has some modifications to both to make them flow better in this order as a possible revision.  I don't think David is convinced this is a good idea, so this is still open to discussion!

I fixed a few other things, but there is an interact in `Burgers` that is still broken, as discussed in #178.  I tried fiddling with this, and there does seem to be an uncaught exception in the last part of `burgers.riemann_solution_interact`.  But I didn't write this gui code and don't really understand how these more complex widgets are supposed to be working. As far as I can see what's done here is identical to what's done in the some of the acoustics widgets that work fine.  So maybe whoever wrote this code (@maojrs ?) can figure out what's going wrong?

### DIFF
--- a/Burgers.ipynb
+++ b/Burgers.ipynb
@@ -42,7 +42,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "In this chapter, we study another scalar nonlinear conservation law: Burgers' equation.  Burgers' equation models momentum transport in a fluid of uniform density and pressure, and it is the simplest equation that captures some key features of gas dynamics.\n",
+    "In this chapter, we study a simple scalar nonlinear conservation law: Burgers' equation.  Burgers' equation models momentum transport in a fluid of uniform density and pressure, and it is the simplest equation that captures some key features of gas dynamics or shallow water flow.\n",
     "\n",
     "To examine the Python code for this chapter, see:\n",
     "\n",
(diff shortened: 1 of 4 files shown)


maojrs:
The interact in Burgers is fine. It just the import statment within burgers.py or burgers_demo.py. I opened an issue a while ago since there were some modifications to Burgers’ notebook that were not neccesary. I opened an issue #185 to discuss them with you.

Regarding te order, it is the same for me. I personally find burgers an easier first equation, and I find it more intuitive, but I might be bias since I have worked a lot with it. 

I am currently on holidays until the 25, but I am happy to do any necessary updates to Burgers when I get back.

ketch:
This looks good and I agree that it's a good idea.

I notice that you added extra blank lines around some equations.  This improves the appearance in the notebook, but adds a paragraph break in the PDF.  I haven't found a solution to this problem.

rjleveque:
Thanks @maojrs, I knew I'd seen something from you about what the problem was but couldn't find it.  I fixed it and closed #185.  

I also put this interact in a separate section and started a new section to discuss the animations that follow, perhaps clearer especially in the pdf that doesn't have the interact.  I considered putting the `hide` tag on the cell with this interact, but decided to leave it in the pdf even though it doesn't show anything interesting.  What do you think?

@ketch: I removed most of the blank lines around displayed equations, I forgot about that issue.

maojrs:
@rjleveque I like the new order. Regarding the interact of the full solution, I also think it is only interesting because of the interactivity. However, it is also ok to leave it in the pdf, so readers become curious to try the online version. I could go either way, but I think keeping it in the pdf would be better.

I also noticed you added a "fig" input parameter to `burgers_demo/multivalued_solution` and `burgers_demo/shock_location`. From my point of view this extra parameter is not necessary, but maybe you added it for a specific reason?

rjleveque:
I added `fig` as a parameter in many demos at one time so that it would keep using the same figure and not produce a long sequence of plots in the notebook when making an animation.  But maybe there's a way to avoid it or it now just works without it.  I can experiment with that some more.

For now I'm going to merge this PR so we can move forward with other things.

maojrs:
I see, I think it should work fine without it since they are simply interacts, like all the other interacts in the notebook that don't have it, but I could be wrong. In any case, it can easily be changed later.

rjleveque:
But some interacts get turned into animations when making the html version, so maybe that's where I had problems...

PR #188, opened 2019-05-01
Add pyclaw simulation at end of SW approx notebook.

ketch:
When I taught from this chapter, I felt that there ought to be a comparison of full FV solutions using the two Riemann solvers at the end.  This PR adds a PyClaw dam-break simulation with Roe and HLLE.  The basic takeaway is that the two solvers give almost the same solution on any reasonable grid, despite the fact that the underlying approximate Riemann solutions are so different.  I think it is important for students to be aware of this.

I'm open for suggestions of other things to include.

### DIFF
--- a/Shallow_water_approximate.ipynb
+++ b/Shallow_water_approximate.ipynb
@@ -886,6 +886,115 @@
     "sw.phase_plane_plot(q_l,q_r,g=1.,y_axis='u',\n",
     "                    approx_states=states['hll'])"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


ketch:
I don't remember now -- is there an easy way to get my widget to show the last frame instead of the first frame in the PDF?

rjleveque:
This worked once I upgraded to v5.6.0 for the `shallow_hlle_1D` Riemann solver.

I think we had the static widgets set up to show two frames, t=0 and some later time.  But I notice in `Traffic_flow` it works for one example but not for the other, so this needs some work still.  I'll try to investigate.

PR #189, opened 2019-06-07
Fix links and figsize

rjleveque:
Fixing a few things I noticed with links and figsize's, including fixes to #186.

I also added brief descriptions to the Index for Part II notebooks, could perhaps be improved.

I am thinking of removing Parts III - VI from the Index since currently this makes a bunch of broken links on the [html version](http://www.clawpack.org/riemann_book/html/Index.html), and instead just have a pointer to a page on github that lists newer notebooks and plans for future Parts.  What do you think?


rjleveque:
Python3 tests fail -- I will rewrite to avoid using `find`.

rjleveque:
I don't know why Python3 fails now. Any ideas how to fix?

I tried splitting Index in two, with a new Index2.  What do you think?

ketch:
>I don't know why Python3 fails now. Any ideas how to fix?

I can't reproduce the problem locally.  I'm trying some things to see if they fix the problem.

>I tried splitting Index in two, with a new Index2. What do you think?

Looks good to me.

rjleveque:
Now passing, will merge in.

PR #190, opened 2019-06-08
Syntax highlight

rjleveque:
I wrote a script to use pandoc to convert each .py file that we link to from notebooks into a syntax-highlighted .html file like this example: http://staff.washington.edu/rjl/misc/burgers.html

This also has the advantage that it always opens the file in the browser, whereas depending on ones browser setting, clicking on a link to a .py file might just download it rather than displaying it.

What do you think? If you like it we need to make similar changes in all notebooks that link to exact solvers.  

We might also want a notebook or index at least of all the code in `utils` for readers who want to examine these.

By the way I put this script in utils and I think we should move other utilities like `make_pdf.py` there eventually so the main directory isn't cluttered up, to make it easier for the reader to find the notebooks.

### DIFF
--- a/Burgers.ipynb
+++ b/Burgers.ipynb
@@ -47,7 +47,7 @@
     "To examine the Python code for this chapter, see:\n",
     "\n",
     " - [exact_solvers/burgers.py](exact_solvers/pandoc/burgers.html)\n",
-    " - [exact_solvers/burgers_demos.py](exact_solvers/pandocburgers_demos.html)"
+    " - [exact_solvers/burgers_demos.py](exact_solvers/pandoc/burgers_demos.html)"
    ]
   },
   {
--- a/Burgers.ipynb
+++ b/Burgers.ipynb
@@ -46,8 +46,8 @@
     "\n",
     "To examine the Python code for this chapter, see:\n",
     "\n",
-    " - [exact_solvers/burgers.py](exact_solvers/burgers.py)\n",
-    " - [exact_solvers/burgers_demos.py](exact_solvers/burgers_demos.py)"
+    " - [exact_solvers/burgers.py](exact_solvers/pandoc/burgers.html)\n",
+    " - [exact_solvers/burgers_demos.py](exact_solvers/pandocburgers_demos.html)"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


ketch:
I like it.  The only drawback is that it is one more bit of maintenance.  I guess we will want a script that regenerates all of these so it's easy to update them.

>By the way I put this script in utils and I think we should move other utilities like make_pdf.py there eventually so the main directory isn't cluttered up, to make it easier for the reader to find the notebooks.

I agree.

rjleveque:
It would be easy to invoke this script from `make_html_on_master.py`, which we use to make the html versions for the web.  But I guess it would also be good to have an automatic way to remake them whenever we change something in utils or exact_solvers for use when running the notebooks.

And I guess we also have to commit and push all these html files if we want people who clone the repository to have them available when running the notebooks.  Either that or have something like a Startup.ipynb notebook that people are supposed to run first to create these files, and anything else we want to do at startup, but I fear people wouldn't do that and would just launch into some notebook and then be disappointed if the links don't work.

rjleveque:
It occurs to me we could modify this module so it has a function that just syntax highlights a specified set of files, and call this at the top of each notebook in a hidden cell to make the highlighted version of the files needed for that notebook.

ketch:
On second thought, this is starting to seem like unnecessary work.  It seems very unlikely that someone will want to read the code but will not be comfortable opening it in a text editor.  This seems especially true if they have actually cloned the repository. I suspect those people would rather see code in their chosen text editor (with syntax highlighting that is particularly familiar to them) than in the browser.

Meanwhile, there is a syntax-highlighted HTML version that's always up-to-date on Github; e.g.

https://github.com/clawpack/riemann_book/blob/master/exact_solvers/burgers.py

If we think that's helpful we could point people to it at the start of the book.

rjleveque:
All true, and we could also add links to github in the notebooks too, e.g.

To examine the Python code for this chapter, see:

 - [exact_solvers/burgers.py](exact_solvers/burgers.py) ... [on github](https://github.com/clawpack/riemann_book/blob/master/exact_solvers/burgers.py)

The two versions could conceivably be out of sync, but presumably not if we don't change these after publishing and/or if they do a 'git pull'.

rjleveque:
Closing this, but maybe will add explicit github links for code as suggested in my last comment.

PR #194, opened 2019-06-14
More edits to Burgers

rjleveque:
I added a bit more to Burgers about characteristics in particular, and the spreading of the rarefaction fan.

I still like having this notebook first, also because it shows the converging characteristics and shock formation right up front while in Traffic_flow it starts with an animation of the shock wave moving and a plot that shows vehicle paths, but characteristics aren't shown until later.

I also added in links to the github pages for the .py files, is this ok?


maojrs:
It looks good to me.

ketch:
Looks good to me too.  Thanks.

PR #195, opened 2019-06-14
Intro edits

rjleveque:
I added the interactive phase plane plots directly to the Introduction, what do you think?

I also added github code link and added a bit of discussion of syntax highlighting.

In this notebook and some others we still use `[Shallow_water.ipynb](Shallow_water.ipynb)` to refer to other chapters, while others were changed to `[Shallow_water](Shallow_water.ipynb)`, and I note some places the `[text]` is now more than just the title name.  So do we prefer the latter?


maojrs:
I like the interactive plots shown directly. Regarding the referencing, I think  `[Shallow_water](Shallow_water.ipynb)` seems better to me, but I don't have a strong opinion.

I think it is a good idea to have a link to github as well as to the code within the repository. Do we want to add github links to all the notebooks?

ketch:
>I added the interactive phase plane plots directly to the Introduction

This reverts what you did in 1210856165339da47b11bdd675e1b4cf1527bc87.  I don't remember the reasons, but I do remember that change was made after some thoughtful discussion.  Does anybody else remember?

maojrs:
I remember also it was changed, but don't remember why. Maybe there was an issue back then when converting into pdf/html.

rjleveque:
I've forgotten now why I removed it originally, maybe it wasn't showing up properly or we just decided it wasn't really appropriate in the Introduction where we haven't yet discussed the phase plane and structure of the solution in much detail.  

I think they might help draw readers in and encourage them to check out the notebooks.

But they do appear in the Acoustics and Shallow_water notebook and I don't have a strong feeling one way or the other about including them here.

ketch:
Okay; it's fine with me.  I just remembered you deciding to remove it before and wanted to make sure we had thought it through.

rjleveque:
@ketch and @maojrs: I cleaned up make_pdf.py and added Instructions_pdf.md.

Do you want to take a quick look and suggest any improvements before I send to SIAM?

maojrs:
@rjleveque it is working beautifully for me. I always had issues trying to run the make_pdf.py, but now it runs flawlessly. Some comments/suggestions:

- I saw you mentioned bookbook, but it might not be a bad idea to mention an official list of requirements in the `Instructions_pdf.md` .
- The second chapter (SW eq with tracer) didn't seem to get the references to other chapters correctly; they are all ??. See page 27, 29 and 30 of the pdf.
- I saw you added in italics descriptions of how everything would look in the interactive book. You probably already thought about it, but it would be a good idea to explain this to the SIAM people or maybe add a little temporary minipreface explaining this and the basic idea of the pdf/interactive book.

PS. Still working on revisions on Part II. I expect to be done soon.

rjleveque:
Thanks @maojrs for catching the issue with cross-referencing chapters. I commented out too much in `make_pdf.py`, but it's now fixed.

I fixed a few other things,
 - switched to siamplain.bst for bibliography
 - updated `Index2` to properly point to the notebooks not in the book
 - added `Instructions_html.md` for making html versions. 

I think this PR is ready to merge and everything is in pretty good shape.

I haven't written up more general instructions yet for installing everything needed to run the notebooks, maybe someone wants to update [the wiki page](https://github.com/clawpack/riemann_book/wiki/Installation) and/or turn it into a `.md` file.

I did just do a PR #199 with a simplified Dockerfile that seems to have everything needed. Please try it out.

ketch:
I did a minimal update of the wiki installation instructions.  Perhaps we should add Clawpack to the requirements.txt?  Then again, most of the book works without it and it does require a fortran compiler.

This looks good to me too.  Thanks for cleaning up all these things.  I'll merge it now.

Issue #196, opened 2019-06-14
Things to do in multiple notebooks


rjleveque:
A list of some things that we need to do still:

 1. Put github links for all `exact_solvers` codes in each notebook?
 1. If we put these in, modify `fix_latex_file.py` so it doesn't generate two footnotes with github links.
 1. Decide whether to use `[Shallow_water.ipynb](Shallow_water.ipynb)` to refer to other chapters, or `[Shallow_water](Shallow_water.ipynb)` and make consistent.
 1. We use subscripts `q_l` some places and `q_\ell` others.  Make consistent.
 1. Add more references to other books or papers?
 1. Improve the looks of `riemann.html` generated by `make_html_bib.sh`.  Currently the entries are put in the order they appear in `riemann.bib`, not in alphabetical order. Also some have first names, others not.
 1. Move utility scripts in `utils` and change links as needed.
 1. Make a set of files documenting how we do things and the tools we use, maybe as `.md` files in a `doc` folder?  Fix the link to these in `Preface.ipynb`.
 1. Fix up the Dockerfile if necessary and make sure everything works on binder.

PR #197, opened 2019-06-14
Edits to Preface and additions to .bib

rjleveque:
WIP, updating some discussion in Preface.  I removed link to Azure since I don't want to maintain that along with binder.  I added some more references to other books, but not at all sure this is the optimal short list of other references to give, so feel free to suggest changes/additions.

I added a few more notes on things still to do in #196.


ketch:
Should we review and merge this?  I'm not sure if the WIP you mention will go into this PR or a subsequent one.

rjleveque:
Still WIP because I was planning to update some of the information about installing and running on different platforms, but please do look at what is here and make other suggestions.

One question: should we mention "electromagnetic waves" in the first sentence since this isn't a system we discuss or even plan to in future notebooks?   Maybe mention some other sort of waves here instead, such as "elastic wave propagation"?

rjleveque:
Merging for now, we can make more updates later..

PR #198, opened 2019-07-05
Directory structure

rjleveque:
I cleaned up the main directory by moving several things to subdirectories.  I updated `make_pdf.py` and `make_html.py` to find things properly.

In particular I made a `phase_plane` directory for those html files and renamed them, and fixed the notebooks that refer to them.  This will make it easier for the html and pdf versions.


rjleveque:
@ketch and @maojrs: I'm trying to clean things up a bit for SIAM copy editors to have a look.  I think we should ask them to look at the two notebooks `Shallow` and `Shallow_tracer` to start.  The `make_pdf.py` file checked in to this PR makes a pdf file of only those two.  Please take a look and see if this works for you.

Note that missing chapters currently generate broken footnotes (e.g. on the first page of the Shallow chapter) but these will go away if all chapters are included.

ketch:
Looks good to me.

PR #199, opened 2019-07-11
Dockerfile that uses clawpack-v5.6.0 from dockerhub

rjleveque:
This Dockerfile is much simpler and loads on binder much more quickly since it is based on the docker image for clawpack-v5.6.0 on dockerhub.

I just tested this again on binder and it seems to work fine.  Try it at:

https://mybinder.org/v2/gh/rjleveque/riemann_book/Dockerfile_v5.6.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,72 +1,33 @@
-FROM ubuntu:16.04
 
-# Install some useful tools + gfortran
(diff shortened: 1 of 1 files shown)


ketch:
Everything seems to be working correctly.  I'll merge this.

Issue #200, opened 2019-07-12
Comments and suggestions for Part II


maojrs:
@rjleveque @ketch I went over the second part, took me a bit longer than expected. I think it is overall in good shape, but of course there are a couple of thinks we could still improve. My comments/suggestions follow: 

**Chapter 11:**
- Sec 11.1, 4th paragraph: State explicitly what are R and \Lambda. Or at least cite Chapter 4, where we first introduce this diagonalization.
- Sec 11.1, end: Should we cite notebooks not in printed version of the book like acoustics in heterogeneous media?
- Before Sec 11.2 and Sec 11.3: maybe add a small paragraph saying that most approximate (if not all) Riemann solvers fall within two categories: linearized Riemann solvers and two (or more) wave solvers.
- Sec 11.3: Mention that in some special cases additional waves can be added to the two-wave solvers (as is the case with the HLLC)

**Chapter 12:**
- General comment: I think this notebook can be improved. It is nice we show the solutions using clawpack, but it will also be nice to show the actual solution of the approximate Riemann solver and maybe compare it with the exact solution. Specially in the case of the HLL, so it is clear it separates into three states. Also, if I am correct the HLL solver doesn't need an entropy fix for the transonic case. If so, maybe organizing the chapter differently could help make this clear. If it is ok with you, I'll give it a shot in improving this Chapter.

**Chapter 13:**
- Sec 13.1.1, first paragraph: Maybe say explicitly that the averages must be chose to satisfy Property 3.
- Sec 13.1.2, after first code section and before examples: Add subsection for Roe solver examples and maybe a brief description before each example.
- Sec 13.2.2 examples, Same as before, maybe add a brief desription before each example: Dam-break, isolated 2-shock, transonic rarefaction
- The interact from the last two code cells in the notebook doesn't work.

**Chapter 14:**
- Second paragraph: When mentioning the Energy is given by the relation 14.4, maybe also add the ideal gas equation of state for reference.
Sec 14.1, first paragraph. Reference as in Chapter not working in pdf. Maybe this is already fixed in latest version of make_pdf.
Sec 14.2 HLLE Solver: Many parts uses HLL instead of HLLE. I know it is basically the same, but maybe somewhat confusing to have the two acronyms. Maybe change all to HLLE since we are using the HLLE choice of wave speed.
Maybe add the HLLC/HLLEC solver as an example of a three wave HLL solver. I am happy to do this.

**Chapter 15:**
[...]

PR #202, opened 2019-07-17
Implement changes suggested by Mauricio.

ketch:
See #200.


rjleveque:
I think in `Approximate_solvers` we need to replace `\wave` with `{\matcal W}` and `\apdq` with `{\cal A}^+` since the macros aren't found (are they supposed to be)?  Elsewhere these are written out explicitly.

I don't know what's going on with travis, but the tests seem to pass.

ketch:
>I think in Approximate_solvers we need to replace \wave with {\matcal W} and \apdq with {\cal A}^+ since the macros aren't found (are they supposed to be)? Elsewhere these are written out explicitly.

I made this work earlier with a plugin, but it's not worth it at this point.  So I made these changes, except I used \cal because \mathcal didn't work.

rjleveque:
Grepping `mathcal` it seems to be used in other notebooks such as `Shallow_water_approximate` and seems to work fine.  (But note I misspelled it in my comment above, so if you copied and pasted, `matcal` wouldn't work.)

But the resulting rendered version seems to be the same with either one, so I don't think it matters.

ketch:
> if you copied and pasted

LOL, yes, I did exactly that.  Must not have been thinking too clearly at the time...

PR #205, opened 2019-11-21
Minor revisions.

ketch:


rjleveque:
Looks good, thanks!

I notice that in some of these early notebooks the `%matplotlib inline` cell and a cell of import statements appears at the top of the notebook, whereas in most later notebooks we have moved all the code cells to after the initial introductory markdown cells.  Should we try to make this more consistent?

ketch:
>I notice that in some of these early notebooks the %matplotlib inline cell and a cell of import statements appears at the top of the notebook, whereas in most later notebooks we have moved all the code cells to after the initial introductory markdown cells. Should we try to make this more consistent?

Okay, I've incorporated that change into this PR.  I'm still working through the chapters, so this is still WIP.

ketch:
In the shallow water chapter, I changed the code so that certain characteristics are plotted by default; this way they show up in the book.  This is important since we discuss how these characteristics show that the solution is unphysical in these two cases.

ketch:
I'd like to go ahead and merge this now, since other chapters are being edited concurrently and it would be good to get these changes to the copy editors.  Please review.

rjleveque:
I made the pdf from this PR before I read through it all the other day, and in general it all looks fine to me, and can be merged as far as I'm concerned.  I did make a few notes of some minor changes in these notebooks, but I can do those as part of my next pass through updating the notebooks.

So if you are done updating for now then once this is merged I'll make a pass in the next couple days.

ketch:
Sounds good.  I am planning to go through the remaining chapters, but it looks like I won't get to that for a week or so at least.  So go ahead and make a pass.

PR #207, opened 2019-12-01
Changed some figures and default t values of some notebooks

maojrs:
Here I added some modifications following Randy's comments.

- The plots in `Euler_compare` now produce plots of the three variables simultaneously. I think this will make the figures in the pdf more interesting. Also the default time was set to a middle value, so they show in the pdf.

- Changed default time of all plots of `Shallow_water_approximate`, so they show in the pdf

- In `Burgers_approximate` I added the time in the title plots and also changed the default t value, so it shows on the pdf.


rjleveque:
Some more suggestions:

1. In the last cell of `Shallow_water_approximate`, maybe add a line to the def'n of `plot_frame`:
```
plt.title('Depth h at t = %.2f' % t)
```
2. In `Euler_approximate`, it would be nice if the plots generated by `compare_solutions` had a title `t = ...` below the "Waves in x-t plane" plot and on top of the density,velocity,pressure plots, but that seems perhaps hard to do and I didn't want to monkey with the code in `utils/riemann_tools.py` that makes these plots.   The time is implicitly indicated by where the dashed line is in the x-t plot, maybe that's enough but it would be nice to indicate more clearly.  Another possibility is to put it in the legend, e.g. "Exact at t=0.2" and "Roe at t=0.2".

3. For the pdf version, if possible it would be nice if only the later time plot appeared and not the one at t=0 since it's rather boring and doesn't show more than can be deduced from the later time, and the pdf file is a bit figure-heavy in these chapters.

3. I also think these 4 plots are too close together and it would be nice to have a bit of a gap between them for more clarity, though I don't feel strongly about that.

5. In `Euler_compare`, for the W-C blast wave, the grid is so fine that plotting symbols with `'-ob'` and `'-or'` obscures the lines and it's impossible to compare them.  Perhaps just use '-b' and '-r'?

6. For the other plots in this notebook the dots are ok, but maybe a bit too large?  Perhaps add `markersize=4` to these plot commands.

maojrs:
Randy, I just incorporated your suggestions. Summary of changes:
-  Added plot.title line to shallow water approximate (point 1).

- Added two new input variables to `make_plot_function` in `utils/riemann_tools`: `vertical_spacing` and `show_time_legend`. The default values of this variables reproduce previous behavior so there should be no issues with other notebooks. `Vertical_spacing` allows for some space between the plots in the vertical layout, made it 0.15 in `euler_approximate`, and it doesn't change the size of the plots. `Show_time_legend`, if true, shows time in the legend of the plot. (point 2 and 4)

- Changed default times of plots in `euler_approximate`. (point 3)

- In `euler_compare`, I played around with the plotting parameters and set up what I think worked the best. I used dashed and continuous line for the WC blast wave problem (I have to say it is an unfortunate name). Slightly different markersize and linewidth for HLL and Roe in all the other examples. (point 5 and 6)

Happy to add more modifications if needed.

rjleveque:
Looks good, thanks!

PR #209, opened 2019-12-04
Merge SIAM copyedits together with recent changes

rjleveque:
The notebooks `Shallow_water` and `Shallow_tracer` were recently returned from SIAM after copy editing. In the meantime, #205 made some changes to these notebooks.  I think I have successfully merged the two into this version, but @ketch should take a look.


ketch:
Thanks for dealing with this.  Looks good.

PR #210, opened 2019-12-04
Some edits to chapters 2,3,4

rjleveque:
Some suggested changes in the first three chapters after the preface.

I just noticed the Preface comes out as Chapter 1 in the pdf, maybe we need to fix this.

### DIFF
--- a/exact_solvers/acoustics_demos.py
+++ b/exact_solvers/acoustics_demos.py
@@ -58,7 +58,6 @@ def decompose_q(p,u,K,rho):
 
     
     axes[1].plot([0,p],[0,u],'k-',lw=3,markersize=8)
-    #axes[1].plot([p],[u],'ko',markersize=8)    
     alpha1 = (Z*u-p)/(2.*Z)
     alpha2 = (Z*u+p)/(2.*Z)
     axes[1].plot([0,-Z*alpha1],[0,1*alpha1], color=colors[0], lw=3)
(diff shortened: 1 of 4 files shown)


ketch:
>I just noticed the Preface comes out as Chapter 1 in the pdf, maybe we need to fix this.

Yes; I didn't find a nice way to automate that so I think it's best to just do it manually.

exact_solvers/acoustics_demos.py
```diff
@@ -40,30 +40,38 @@ def decompose_q(p,u,K,rho):
...
     axes[0].legend(['$r_1$','$r_2$'],loc=3)
     axes[0].plot([0,0],[-2,2],'--k',alpha=0.2)
     axes[0].plot([-2,2],[0,0],'--k',alpha=0.2)
+    axes[0].set_xlabel('p')
+    axes[0].set_ylabel('u')
 
     
-    axes[1].plot([0,p],[0,u],'k',lw=3)    
+    axes[1].plot([0,p],[0,u],'k-',lw=3,markersize=8)
+    #axes[1].plot([p],[u],'ko',markersize=8)
```
ketch: I'm guessing you meant to eventually delete this line and the other one also commented out just below.

rjleveque: Yes, thanks.  Now done.

rjleveque:
This is ready to merge if it looks ok to you.  I'll do separate PRs for changes to other chapters.

PR #211, opened 2019-12-04
Some edits to chapters 5,6

rjleveque:
Minor changes in `Burgers`, including introducing the term "Lax Entropy Condition" for the converging characteristic variant.

More substantial changes in `Traffic_flow`, including adding more discussion of a few things and making it flow better now that `Burgers` comes first.  

Also I changed the way 3 figures from the `./figures` directory are displayed.  The old version created floating figures with figure numbers and captions in the pdf version.  But we don't refer to these figure numbers anywhere so they ended up not well aligned with the discussion of them. Now I use `Image` to display them inline the way all other figures do, so no more floats or figure numbers.  Also the `width` parameter is used to make  them a bit smaller in the notebook.  I was hoping this might carry over to the pdf, where I think they also come out too big, but it doesn't automatically.  However, by editing the latex, e.g. from
```
\adjustimage{max size={0.9\linewidth}{0.9\paperheight}}{combined_files/combined_16_0.png}
```
to
```
\adjustimage{max size={0.5\linewidth}{0.5\paperheight}}{combined_files/combined_16_0.png}
```
they come out looking pretty good, IMHO.

A disadvantage with this approach is that we cannot "hide" the input cell calling `Image` or it doesn't show up at all in the pdf, but this could also be edited out from the latex at the end if desired.

What do you think?

Also, should the second of these figures (`figures/shock_diagram_traffic_a.png`) have the dashed vertical line?  That seems to be needed only in the third figure with the moving shock (`figures/shock_diagram_traffic_b.png`)


ketch:
These changes all look good to me.  I guess that in the pdf you sent by email, you edited out the "Image" calls?  We could probably modify bookbook to do that automatically, but it has been a long time since I dove into that stuff so it would likely take me hours to figure it out again and so may be simpler to handle manually (or with a shell script).

You are right that the first figure would be better without the dashed line.  I'll open a PR to change that.

rjleveque:
Yes, I did edit out the Image cells.  I wouldn't bother trying to automate that.  We can either do it in the final edits (e.g. as we add the \index commands), or it's not so bad if they show up in the book.

PR #212, opened 2019-12-05
Some edits to chapters 7, 10

rjleveque:
Some changes:

`Nonconvex_scalar`:

- Removed the section on Linear Degeneracy as out of place (and expanded the discussion of this in `Euler` and `Shallow_tracer`)
- Improved the discussion some place, in particular of convex hull

`Euler`:

- Added titles to some plots generated in `exact_solvers` codes, and changed interactive Riemann solver for Euler so it has different data than Sod, and so it's clear what these data are in the call to `euler_demo1`, to improve what shows up in the pdf file.
- Changed lots of **bold** text to *italics* to better match what we do in other chapters, where bold is rarely used.
- Reorganized definitions of genuinely nonlinear and linearly degenerate to discuss more in more generality.
- Added a bit of discussion to the last two sections on vacuum states.

One question in `Euler`: do we really want to write out the full 3D Euler equations, as we do in this notebook?  We don't do this for any other problem, although we could for acoustics, SWE, etc.  Maybe it's useful to see but perhaps just distracting since we don't discuss it further.


Euler.ipynb
```diff
@@ -342,7 +346,7 @@
     "\n",
     "1. Define a piecewise function giving the middle state velocity $u_m$ that can be connected to the left state by an entropy-satisfying shock or rarefaction, as a function of the middle-state pressure $p_m$.\n",
     "2. Define a piecewise function giving the middle state velocity $u_m$ that can be connected to the right state by an entropy-satisfying shock or rarefaction, as a function of the middle-state pressure $p_m$.\n",
-    "3. Use an iterative solver to find the intersection of the two functions defined above.\n",
+    "3. Use an nonlinear rootfinder to find the intersection of the two functions defined above.\n",
```
ketch: I guess this should be "a nonlinear rootfinder".

ketch:
These changes look good to me.  I think we should remove the 3D Euler equations, since we don't even talk about shear waves in the Riemann solution, and interested readers can go to the other books we cite.

Euler.ipynb
rjleveque: Yes, thanks.  I changed this wording for consistency with Shallow_water.

rjleveque:
I removed 3D Euler and slightly reworked the paragraph around the 1D equations, and eliminated one section heading.

PR #213, opened 2019-12-05
Update first traffic figure for shock location (remove dashed line).

ketch:
As suggested in #211.


rjleveque:
Looks good, thanks!

PR #215, opened 2019-12-08
New material added to Approximate_solvers

rjleveque:
I added quite a bit to `Approximate_solvers`, let me know what you think.

I thought it would be good to say a bit more about FV methods, Godunov in particular, and how the Riemann solver is used. And I rewrote it to remove the fluctuations amdq and apdq since we never use these later and people often find that notation confusing.  Instead I allude to them in the context of nonconservative problems but focus more on flux-differencing form and interface fluxes.

I'm still going through the following chapters to make sure notation is consistent with these changes, and have some other things to fix in those.


Approximate_solvers.ipynb
```diff
@@ -36,6 +36,51 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Finite volume methods\n",
+    "\n",
+    "We give a short review of Riemann-based finite volume methods to illustrate what is typically needed from a Riemann solver in order to implement such methods. In one space dimension, a finite volume approximation to the solution $q(x,t_n)$ at the $n$th time step consists of discrete value $Q_j^n$, each of which can be viewed as approximating the cell average of the solution over a grid cell $x_{j-1/2} < x < x_{j+1/2}$ for some discrete grid.   The cell length $\\Delta x_j = x_{j+1/2} - x_{j-1/2}$ is often uniform, but this is not required.  Many methods for hyperbolic conservation laws are written in *conservation form*, in which the numerical solution is advanced from time $t_n$ to $t_{n+1} = t_n + \\Delta t_n$ by the explicit formula\n",
```
ketch: "discrete value" -> "discrete values"

Approximate_solvers.ipynb
```diff
@@ -36,6 +36,51 @@
...
+    "\n",
+    "\\begin{align}\\label{FVupdate}\n",
+    "Q_j^{n+1} = Q_j^n - \\frac{\\Delta t_n}{\\Delta x_j} (F_{j+1/2}^n - F_{j-1/2}^n),\n",
+    "\\end{align}\n",
+    "\n",
+    "for some definition of the *numerical flux* $F_{j-1/2}^n$, typically based on $Q_{j-1}^n$ and $Q_j^n$ (and possibly other nearby cell values at time $t_n$).\n",
+    "\n",
+    "Dividing (\\ref{FVupdate}) by $\\Delta t_n$ and rearrraning, this form can be viewed as a discretization of $q_t + f(q)_x = 0$, provided the numerical flux is *consistent* with the true flux $f(q)$ in a suitable manner. In particular if the $Q_i$ used in defining $F_{j-1/2}^n$ are all equal to the same value $\\bar q$, then $F_{j-1/2}^n$ should reduce to $f(\\bar q)$.\n",
```
ketch: rearranging

Approximate_solvers.ipynb
```diff
@@ -36,6 +36,51 @@
...
+   "source": [
+    "### Godunov's method\n",
+    "\n",
+    "Trying to compute numerical approximations to hyperbolic problems with strong shock waves is challenging because of the discontinuities in the solution --- classical interpretations of $(F_{j+1/2}^n - F_{j-1/2}^n)/\\Delta x_j \\approx \\partial f/ \\partial x$ break down, oscillations near discontinuities often appear, and methods can easily go catastrophically unstable.\n",
+    "\n",
+    "The landmark paper of Godunov <cite data-cite=\"godunov\"><a href=\"riemann.html#godunov\">(Godunov 1959)</a></cite> was the first to suggest using the solution to Riemann problems in defining the numerical flux: $F_{j-1/2}^n$ is obtained by evaluationg $f(Q_{j-1/2}^*)$, where $Q_{j-1/2}^*$ is the Riemann solution evaluated along the ray $x/t = 0$ after the Riemann problem is solved between states $q_\\ell=Q_{j-1}^n$ and $q_r=Q_j^n$. If the numerical solution is defined as a piecewise constant function with value $Q_j^n$ in the $j$th cell, then the exact solution takes this value along the cell interface $x_{j-1/2}$ for sufficiently small later times $t > t_n$ (until waves from other cell interfaces begin to interact).\n",
```
ketch: evaluating

ketch: It seems necessary here to clarify that the Riemann problem is solved in a coordinate system where $x_{j-1/2}=0$.

Approximate_solvers.ipynb
```diff
@@ -36,6 +36,51 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### High-resolution methods\n",
+    "\n",
+    "In spite of its robustness, Godunov's method is seldom used as just described because the built-in *numerical viscosity* that gives it robustness also leads to very smeared out solutions, particularly around discontinuities, unless a very fine computational grid is used.  For the advection equation, Godunov's method reduces to the standard \"first-order upwind\" method and it general it is only first order accurate even on smooth solutions.\n",
```
ketch: "it general" -> "in general"

Approximate_solvers.ipynb
```diff
@@ -36,6 +36,51 @@
...
+    "\n",
+    "A wide variety of higher order Godunov-type (i.e., Riemann solver based) methods have been developed.  One approach first reconstructs better approximations to the solution at each time from the states $Q_j^n$, e.g. a piecewise polynomial that is linear  or quadratic in each grid cell rather than constant, and then uses the states from these polynomials evaluated at the cell interfaces to define the Riemann problems.  Another approach, used in the Clawpack software, is to take the waves that come out of the Riemann solution based on the original data to also define second order correction terms.  In either case some *limiters* must generally be applied in order to avoid nonphysical oscillations in solutions, particular when the true solution has discontinuities.  There is a vast literature on such methods, see for example many of the books cited in the [Preface](Preface.ipynb).  For our present purposes the main point is that an approximate Riemann solver is a necessary ingredient in many methods that are commonly used to obtain high-resolution approximations to hyperbolic PDEs."
```
ketch: The first approach mentioned here is also used in Clawpack (SharpClaw), so maybe we should rephrase this.

ketch: "particular" -> "particularly"

ketch: "methods," -> "methods;"

Approximate_solvers.ipynb
```diff
@@ -47,55 +92,74 @@
...
+    "\n",
+    "The Riemann solution gives a resolution of the jump $\\Delta q = (q_r - q_\\ell)$ into a set of propagating waves.  In both of the approaches described below, the approximate Riemann solution consists entirely of traveling discontinuities, i.e., there are no rarefaction waves in the approximate solution, although there may be a discontinuity that approximates such a wave.  One should rightly worry about whether the approximate solution generated with such a method will satisfy the required entropy condition and end up with rarefaction waves where needed rather than entropy-violating shocks, and we address this to some extent in the examples below.  It is important to remember that we are discussing the approximate solver that will be used *at every grid interface* in every time step, and the numerical viscosity inherent in the numerical method can lead to rarefactions in the overall numerical approximation even if each approximate Riemann solution lacks rarefactions.  Nonetheless some care is needed, particularly in the case of *transonic* rarefactions, as discussed further below.\n",
```
ketch: "below" -> "in the following chapters"?

Approximate_solvers.ipynb
```diff
@@ -47,55 +92,74 @@
...
     "\\end{align}  \n",
     "This can be solved for the middle state to find  \n",
     "\\begin{align}  \\label{AS:middle_state}\n",
-    "q_m = \\frac{f(q_r) - f(q_l) - s_2 q_r + s_1 q_l}{s_1 - s_2}.\n",
+    "q_m = \\frac{f(q_r) - f(q_\\ell) - s_2 q_r + s_1 q_\\ell}{s_1 - s_2}.\n",
     "\\end{align}  \n",
     "It remains only to specify the wave speeds, and it is in this specification that the various two-wave solvers differ.  In the following sections we briefly discuss the choice of wave speed for a scalar problem; the choice for systems will be elaborated in subsequent chapters.\n",
     "\n",
-    "In addition to the names and references provided below, let us mention that this class of solvers is an ingredient in the so-called *central schemes*.  Due to the extreme simplicity of two-wave solvers, the resulting central schemes are often even referred to as being \"Riemann-solver-free\"."
+    "Typically $s_1 < 0 < s_2$ and so the intermediate state we need is $Q^* = q_m$ and $F^* = f(Q^*)$.  However, in some cases both $s_1$ and $s_2$ could have the same sign, in which case $F^*$ is either $f(q_\\ell)$ or $f(q_r)$.\n",
```
ketch: Are there approximate two-wave solvers where this can actually occur?  The ones I can think of assume the waves travel in opposite directions.

ketch:
This looks very good to me, modulo a few minor comments above.  This will make the book significantly more self-contained and provides better motivation for all of Part II.

rjleveque:
Thanks for reviewing.  I'll make these edits.

Approximate_solvers.ipynb
rjleveque: I tried to word this part as solving the standard Riemann problem (as discussed up to now) with states $q_\ell$ and $q_r$, and then the the state along $x/t = 0$ in this solution is the one that appears along the cell edge on the finite volume grid.  I tried rewording this a bit, see if it is any clearer now.

Approximate_solvers.ipynb
rjleveque: If you apply HLL to a supersonic Riemann problem for Euler (or supercritical for SWE) then the eigenvalues could all have the same sign in both states so the formula at the bottom of this notebook would give the same sign for both speeds.

Approximate_solvers.ipynb
rjleveque: I reworded to mention wave propagation methods rather than Clawpack.

Approximate_solvers.ipynb
ketch: I agree that the wave speeds in the true Riemann solution can all have the same sign. But I don't know of an approximate two-wave solver in the literature that allows two waves with the same sign.  The formula at the bottom of this notebook is for a scalar problem, not for systems.  From my understanding of the HLL solver, in the supersonic/supercritical case, the other wave would be given speed zero.

ketch: In my mind, the idea of a two wave solver is to capture all the effects of waves going in a given direction through a single consolidated wave (basically a fluctuation).  Of course, one could design a different kind of two-wave solver; I just haven't seen it.

rjleveque: I pointed to the wrong scalar equations, but see the expressions for $s_1$ and $s_2$ in `Shallow_water_approximate` or `Euler_approximate`.  Won't these both be the same sign in the supercritical cases?

ketch: Hmmmm, you're right. I guess this is a misconception on my part, although I'm certain I have seen some papers where they take the min/max with zero for those speeds.  Anyway, I withdraw my objection to this.

PR #216, opened 2019-12-09
Edits to final chapters

rjleveque:
Some edits to `Burgers_approximate`, `Shallow_water_approximate`, `Euler_approximate`, and `Euler_compare`.

Question: Should we merge `Euler_approximate`, and `Euler_compare` into a single notebook? 
Or split `Shallow_water_approximate` into two?

At the moment the two are handled differently, since `Shallow_water_approximate` has both discussion of the single Riemann solution and PyClaw examples using approximate solvers, whereas for Euler we spit them up.


Burgers_approximate.ipynb
```diff
@@ -198,7 +207,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The solutions obtained for the shock wave and for the first rarefaction wave are good approximations of the true solution.  In the case of the transonic rarefaction, however, we see that part of the rarefaction has been replaced by an entropy-violating shock.  At the end of this chapter we will show how to apply an *entropy fix* so that the solver gives a good approximation also in the transonic case."
+    "The solutions obtained for the shock wave and for the first rarefaction wave are good approximations of the true solution.  In the case of the transonic rarefaction, however, we see that part of the rarefaction has been replaced by an entropy-violating shock.  Refining the grid in this numerical solution would not help, the method converges to a weak solution that does not satisfy the entropy condition.  *In the live notebook you can try adjusting $N$ in the cell above to investigate this.*\n",
```
ketch: "help," -> "help;"

Shallow_water_approximate.ipynb
```diff
@@ -15,8 +15,24 @@
...
+    "\n",
+    " - [exact_solvers/shallow_water.py](exact_solvers/shallow_water.py) ...\n",
+    "   [on github.](https://github.com/clawpack/riemann_book/blob/master/exact_solvers/shallow_water.py)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We investigated the exact solution of the Riemann problem for this system in [Shallow_water.ipynb](Shallow_water.ipynb).  found that the system has two genuinely nonlinear characteristic fields, so the solution consists of two waves, each of which may be a shock or a rarefaction.\n",
```
ketch: I think the word "We" was accidentally deleted here.  Maybe just end the previous sentence with a comma and insert "and"?

ketch:
Looks good!  When we wrote `Euler_compare`, we didn't have the PyClaw example at the end of the SW chapter.  I don't like splitting the SW example into its own chapter, since it's just the one example without a lot of discussion.  Merging the two Euler chapters makes a rather long notebook (due mostly to the long code blocks for the PyClaw simulations) but I think it's not longer than some of the other chapters, so that seems fine.

maojrs:
I also think it would be fine to join them . However, an alternative idea would be to change the name of `Euler_compare` to something like `Comparison of approximate solvers` and move the last example of `Shallow_water_approximate` into the comparison notebook. This way, we keep a chapter just aimed in the comparison between different solvers, and avoid having a very large `Euler_approximate` chapter.

rjleveque:
I like the idea of moving the SWE example to the last chapter so all the PyClaw code is together. I'll try that and see how it looks.

rjleveque:
I combined the examples and also rearranged the order of the Euler examples and code.  I renamed the last notebook to `FV_compare`

If this looks ok, there are other places this name change needs to be made, in particular in `Index.ipynb`, `make_html.py`, and `make_pdf.py`.

ketch:
Looks good to me.  I would say go ahead and make the name change.

maojrs:
Randy, I updated the plotting style in the `FV_compare` notebook of the SW example to match how it looks in the other examples. I also added the momentum in the same plot. If you like how it looks, I made a pull request to your branch.

PR #221, opened 2020-01-08
in Burgers, the import statements appeared twice

rjleveque:
I removed from both places and relocated to where they seem least disruptive, before the "shock formation" section.

### DIFF
--- a/Burgers.ipynb
+++ b/Burgers.ipynb
@@ -8,38 +8,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "tags": [
-     "hide"
-    ]
-   },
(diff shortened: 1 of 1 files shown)


rjleveque:
No idea at the moment why the travis tests fail on `Euler.ipynb`.

ketch:
It looks like you removed `matplotlib inline` and another line about figure formatting -- was that intentional?

rjleveque:
I think the matplotlib inline is still in there in one place.  It used to be in there twice.

I think this is the new version:

https://github.com/clawpack/riemann_book/blob/2844dd68034b87da0eef01c51b79e0faaf716132/Burgers.ipynb

ketch:
You're right; my mistake.

PR #222, opened 2020-01-15
Change notebook version number.

ketch:
This is a quick fix to deal with a bug related to recent nbformat updates; see https://github.com/jupyter/nbformat/issues/161.  It should fix all the recent failures we've been getting.  Fingers crossed.


ketch:
It seems that this fixes the problems we've been seeing.

rjleveque:
Thanks @ketch!  I experimented a bit and simply running a notebook and saving a new version doesn't change this version number, but if @maojrs or I make new notebooks in the future we'll have to make sure we have the updated notebook version with nbformat 4.   (Which I installed and things seem to work fine.)

PR #223, opened 2020-02-14
fix float index in riemann_tools

alexfikl:
Using float steps in `np.linspace` was deprecated in numpy 1.12 and removed in 1.18 with [this commit](https://github.com/numpy/numpy/commit/f4dfe833e3e037bb69113f7250fad3699f918cfc).

With this change `make_pdf.py` runs and produces the pdf.


rjleveque:
Thanks for catching that.

PR #224, opened 2020-02-18
Remove sympy dependency

rjleveque:
sympy is only used if verbose==True and I don't think we use this in the notebooks, so I suggest we move the import into this `if verbose` block.


ketch:
Good idea!

PR #231, opened 2020-07-05
Update README.md for SIAM book

rjleveque:
Add more links now that the SIAM book is published, and various minor improvements.


ketch:
This looks great to me; thanks!

rjleveque:
I'll merge this and then we can still try also adding the index directly to see how that looks.

PR #232, opened 2020-07-07
Update README.md

rjleveque:
Add Contents section copied from Index.ipynb

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The recommended way to fully experience the book is by running the Jupyter noteb
  - [Locally, installed manually](#installation)
  - [In the cloud with Binder](#binder)
  
- Start from the table of contents given in the notebook [Index.ipynb](Index.ipynb).
+ Start from the table of contents given in the notebook [Index.ipynb](Index.ipynb), which is also [shown below](#contents).
  The notebook [Index2.ipynb](Index2.ipynb) lists some notebooks that are not 
  in the SIAM book, some of which are still under development.  Additional notebooks may appear in the future.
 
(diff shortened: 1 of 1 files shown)


ketch:
This seems fine; I find that occasionally Github is slow to render notebooks or even fails.  That mostly happens with large notebooks so probably wouldn't affect the index, but this just potentially makes it easier for some people to access.

ketch:
Now that I merged it I just realized that it doesn't include the hyperlinks -- was that intended?  I think we should include the links; if you agree then I can do that.

maojrs:
I think we should definitely include the links.

rjleveque:
I intentionally didn't include the links because linking to the notebook
has the same problem David mentioned that sometimes they don't load (and
this happened for me even with Index.ipynb the other day).  And if it does
load all they see is the input cells stripped of output, and so not so
useful and potentially confusing to those not familiar with notebooks.

We could instead link to the html versions, but that's potentially
confusing in other ways.  Since there are other links in the README to the
rendered notebooks and instructions on how to open notebooks in a way that
they can execute them, I thought the purpose of putting the index here was
simply so people can more easily see what's in the book without having to
open anything.  But I'm happy to be convinced otherwise.

 - Randy

maojrs:
Ohh I see. In that case I think it is also ok as it is now.

ketch:
What if I change the chapter names in the TOC at the bottom of the README to be the actual notebook file names (in backticks), so it's completely clear which notebook is referred to?

rjleveque:
@ketch - Yes, that sounds like a great solution.

ketch:
Okay, I went ahead and just edited the README directly.

rjleveque:
Looks good, thanks!

PR #238, opened 2021-09-29
Force `nsteps` to be `int` in call to `linspace`

mandli:
I just tried to run the book and ran into NumPy complaining about `nsteps` not being an `int` in a call to `linspace`.  This fixes it although I am not entirely sure why `nsteps` has become a `float`.

### DIFF
--- a/utils/riemann_tools.py
+++ b/utils/riemann_tools.py
@@ -565,7 +565,7 @@ def compute_riemann_trajectories(states, s, riemann_eval, wave_types=None,
 
     nsteps = 200.
     dt = 1./nsteps
-    t_traj = np.linspace(0,1,nsteps+1)
+    t_traj = np.linspace(0,1,int(nsteps+1))
     q_old = riemann_eval(xx/1e-15)
     for n in range(1,len(t_traj)):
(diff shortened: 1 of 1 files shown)


rjleveque:
@mandli: Your PR seems to be based on an out-of-date version of the code.  The version of this file in `master`  has `nsteps = 200` as an integer, along with some other changes. See line 572 of https://github.com/clawpack/riemann_book/blob/master/utils/riemann_tools.py
and compare to line 566 of your file.

I'm confused about why this PR is only showing one change when I think there are some other changes between your version and master's.

mandli:
I newly cloned this today so I am not sure what's going on.  Obviously if the easy way to get rid of this problem is to remove a decimal point that is the solution but if I cloned the head of master and I got something that broke it might be good to figure that out.

rjleveque:
I figured it out, I think...

I suspect you cloned from `rjleveque/riemann_book` rather than from `clawpack/riemann_book`.  I just noticed that my fork was way out of date on github, and that version did have the decimal point!

I just pushed updates to `rjleveque`.

mandli:
Ah, I did not even notice, thanks for catching that.

PR #245, opened 2023-01-22
Avoid plotting too-short arrows

ketch:
This removes a visual bug, shown in the attachment.  In this visualization, short arrows go past where they should end.  And one arrow is plotted with length zero, which seems to result in a vertical line.  This PR just avoids plotting any arrows until there is enough space for them in the plot (second screenshot).

Screen Shot 2023-01-22 at 8 52 36 AM
Screen Shot 2023-01-22 at 8 56 16 AM


rjleveque:
Looks good thanks!

Presumably we should merge this and some other recent changes into the FA16 branch too, the one linked from the book.

